Issue #9, opened 2018-03-12
Develop a comprehensive editor for additional properties of labeled nodes


gaurav:
"Additional properties" are how we annotate labeled phylogeny nodes: it consists of a dictionary in the JSON file that associates node labels with the RDF properties associated with them. We primarily use them to store represented taxonomic units and phyloreferences we expect to match at a particular node. However, it can be used to annotate nodes with information on the specimens or genes included in that node, and might be one way of linking our work directly with biodiversity data.

As of @ad9697e9c19 (Mar 12, 2018), some groundwork has been laid for this feature through providing methods for getting a list of all properties associated with a phylogeny and a table-based visualization and editing of these properties. However, a lot more work is necessary to make this fully usable: different types of properties need to be visualized in different ways, filters are necessary to identify particular nodes and properties, and the ability to add new properties to a labelled node is necessary. Finally, this can only edit labeled nodes, so the ability to label new nodes (#8) would be useful.

For now, I'll leave this feature in its current incomplete feature, and fill this issue as a reminder to complete it when we have more time or it becomes more urgent.

gaurav:
While adding additional data to labeled nodes could be useful, this is probably outside the scope of the Curation Tool. If we need to support additional kinds of specifiers (e.g. as in #20), then we will add the UI fields for those terms on a case-by-case basis rather than building a single general system for adding RDF facts to all nodes. Therefore, I don't think we should work on this feature any longer.

As I mentioned previously, there is already a simple additional property table included in the Curation Tool. I'll use this issue to track removing that table, and then I'll close this issue.

PR #10, opened 2018-03-12
Add support for matching specifiers

### LINKED ISSUE #6: Add support for matching specifiers to phylogenetic nodes
The simplest way to implement matching in the Curation Tool would be match taxonomic units to nodes in Javascript. This requires:
 - [ ] An additional viewer and editor for adding additional properties to nodes.
 - [x] A set of matching algorithms that can match taxonomic units using scientific names, specimen identifiers or external references.
 - [ ] An interface to quickly identify and help debug non-matching specifiers. This might require suppressing seemingly correct matches.

gaurav:
Adds support for editing taxonomic units associated with phylogeny nodes and for matching these taxonomic units with those present in specifiers, allowing specifier matching to be visualized within this interface. This pull request also contains the first round of development towards a comprehensive labeled node property editors (#9), which will be completed when this feature is needed. Adding support for taxonomic unit matching allowed us to eliminate the no-longer-needed "additionalLabels" feature, which has been removed from example files and is no longer supported by the Javascript code.

Closes #6. After review, this pull request will be merged.

### DIFF
--- a/js/curation-tool.js
+++ b/js/curation-tool.js
@@ -41,7 +41,11 @@ var vm = new Vue({
         selected_tunit_list: undefined,
         selected_tunit: undefined,
 
-        // Taxonomic unit editor modal tweaks
+        // Phylogeny view modes
+        phylogeny_newick_mode_for: undefined,
+        phylogeny_annotations_mode_for: undefined,
+
+        // Taxonomic unit editor modal elements
         tunit_editor_target_label: "(unlabeled)",
 
         // Display the delete buttons on the specifiers.
--- a/js/curation-tool.js
+++ b/js/curation-tool.js
@@ -541,9 +541,131 @@ var vm = new Vue({
             }
 
             return tunits;
+        },
+
+        // Taxonomic unit matching!
+        get_all_node_labels_matched_by_specifier: function(specifier) {
+            let node_labels_with_prefix = [];
+
+            let prefix = "phylogeny_";
+            let prefix_count = 0;
+            for(let phylogeny of this.testcase.phylogenies) {
+                for(let node_label of this.get_node_labels_matched_by_specifier(specifier, phylogeny)) {
+                    node_labels_with_prefix.push(prefix + prefix_count + ":" + node_label);
+                }
+                prefix_count++;
+            }
+
+            return node_labels_with_prefix;
+        },
+
+        get_node_labels_matched_by_specifier: function(specifier, phylogeny) {
+            let local_vm = this;
+
+            return local_vm.get_node_labels_in_phylogeny(phylogeny).filter(function(node_label) {
+                return local_vm.does_specifier_match_node(specifier, phylogeny, node_label);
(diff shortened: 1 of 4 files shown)


js/curation-tool.js
```diff
@@ -413,10 +464,308 @@ var vm = new Vue({
...
+
+    return false;
+}
+
+/**
+ * do_scnames_match_by_binomial_name(scname1, scname2)
+ *
+ * Test whether two scientific names match on the basis of their binomial names.
+ */
+function do_scnames_match_by_binomial_name(scname1, scname2) {
```
hlapp: Functions that return a boolean generally won't stand on their own when called. Hence, giving them names that constitute a full sentence make not only for awkward reading when they're invoked (`if (do_sth_match_by_sth())`) but also can be easily misread as an action (`if (do_something)`).

Maybe the better model here is having a TU class and a ScientificName class, and then give them methods such as `tu.matches_TU()` and `tu.matches_scientificName`.

BTW camelCase is generally far more common in JS than snake_case. I'd suggest you use the style that's prevalent in the respective language. I.e., don't make JS look like Python.

js/curation-tool.js
```diff
@@ -413,10 +464,308 @@ var vm = new Vue({
...
+
+    return false;
+}
+
+/**
+ * do_tunits_match_by_binomial_name(tunit1, tunit2)
+ *
+ * Test whether two Tunits match on the basis of their binomial names.
+ */
+function do_tunits_match_by_binomial_name(tunit1, tunit2) {
```
hlapp: See above.

js/curation-tool.js
```diff
@@ -413,10 +464,308 @@ var vm = new Vue({
...
+
+    return false;
+}
+
+/**
+ * do_tunits_match_by_specimen_identifier(tunit1, tunit2)
+ *
+ * Test whether two Tunits match on the basis of their specimen identifiers.
+ */
+function do_tunits_match_by_specimen_identifier(tunit1, tunit2) {
```
hlapp: See above.

js/curation-tool.js
```diff
@@ -413,10 +464,308 @@ var vm = new Vue({
...
+    return false;
+}
+
+/**
+ * get_specimen_identifier_as_urn(specimen)
+ *
+ * Given a specimen, return a specimen identifier as a URN in the form:
+ *   "urn:catalog:" + institutionCode (if present) + ':' + collectionCode (if present) + ':' + catalogNumber (if present)
+ */
+function get_specimen_identifier_as_urn(specimen) {
```
hlapp: See above.

gaurav: Issue #35 covers the Specimen class as well. I think this name is okay for now, but please let me know if you have a suggestion for a better name! When I convert it into camelCase, I plan to rename it to `getSpecimenIdentifierAsURN()`, which should be a little clearer.

Issue #32, opened 2018-04-04
Add support for testing multiple phyloreferences resolving to the same node


gaurav:
At the moment, a curator cannot indicate that a single node (whether internal or terminal) is expected to have multiple phyloreferences resolve to it. Our data model supports this by allowing labeled nodes to have multiple `expectedPhyloreferenceNamed` values.

We can fix this by displaying the nodes expected to be matched on each phylogeny in the phyloreference panel (which is also part of #23). From the user's perspective, each phyloreference panel would have an expected labeled node. If multiple phyloreferences expected to resolve to the same node, they would all render correctly. Internally, the model would be implemented in the opposite direction: the labeled nodes would have information on which phyloreferences are expected to resolve to them. 

This may seem counterintuitive, but I think it makes more sense than annotating the phyloreferences: the phyloreference don't need to know which phylogenies it may be resolved against, but the curated phylogenies should know where the expectations are for particular phyloreferences to resolve against them.

gaurav:
This has now been implemented as part of #34, but there is an outstanding bug: although the system works correctly, changing the expected node isn't immediately reflected in the user interface until the user selects another phyloreference and then returns to the changed one. I'll close this issue once that bug has been fixed.

PR #33, opened 2018-04-04
Added support for changing node labels on phylogenies

### LINKED ISSUE #27: Add support for renaming nodes on phylogenies
This is a particular challenge because we need to change the nodes within those processed by PhyloTree, and then we need to reconvert the whole tree back into Newick to store in the JSON file. However, it will be a lot simpler to rename node labels than to edit the taxonomic units associated with a node, so this will be very useful.

gaurav:
This pull request adds support for changing node labels on phylogenies. It does this by modifying the Phylotree internal model, asking the Phylotree library to convert its internal model into Newick, and then overwriting the Newick string in the PHYX model with the updated Newick string.

Closes #27. This pull request does not support multiple phyloreferences expected to resolve on the same node, since those expectations are expressed as internal node labels only (#32).

Will be merged into master after review.


js/curation-tool.js
```diff
@@ -356,7 +356,7 @@ var vm = new Vue({
             if(phylotree === undefined)
                 return "()";
 
-            return phylotree.get_newick(function(n) { return n; });
+            return get_newick_string_from_phylotree(phylotree);
```
hlapp: It looks like we should extend the phylotree object for our purposes? If that's not filed as an issue yet, it should be.

js/curation-tool.js
```diff
@@ -897,6 +897,29 @@ function load_json_from_local(file_input) {
     fr.readAsText(file);
 }
 
+/**
+ * get_newick_string_from_phylotree(phylotree)
+ *
+ * Given a Phylotree object, convert it into a Newick string.
+ */
+function get_newick_string_from_phylotree(phylotree) {
```
hlapp: JavaScript makes it actually very easy to replace methods with customizations for applications:
```js
let getNewick = tree.get_newick;
tree.get_newick = function(annotFunc) {
    return getNewick.call(this, annotFunc) + ";";
};
```
And afterwards you treat phylotree.get_newick as having the behavior you desire (because it actually does). 

I'm using `tree` here and not`phylotree` because you would do this after the object is created (which I think is in `render_tree()`.

JavaScript uses objects and inheritance by prototyping. You can replace any method at any time with anything you wish.

gaurav:
Thanks, Hilmar! I've filed an issue with Phylotree at https://github.com/veg/phylotree.js/issues/61

PR #47, opened 2018-04-15
Addition field for documenting that a single node resolves to multiple phyloreferences

### LINKED ISSUE #32: Add support for testing multiple phyloreferences resolving to the same node

gaurav:
Allows the curator to document that they expect a single node to resolve to multiple phyloreferences by selecting a matching labeled node from the phyloreferencing panel. I've think this feature might prove necessary to Curation Tool v0.1, since this expectation needs to be documented on many phylogenies.

Closes #32.

**Do not merge**: still in development, and has at least one outstanding bug: changing the expected node resolution works fine, but does not update the display.

### DIFF
--- a/examples/hillis_and_wilcox_2005.json
+++ b/examples/hillis_and_wilcox_2005.json
@@ -5,7 +5,7 @@
     "@type": "testcase:TestSuiteFromPublication",
 
     "curator": "Gaurav Vaidya <gaurav@ggvaidya.com>",
-    "comments": " - The Laurasiarana expected node is not found in the TreeBase tree, so I've placed it where I think it should go.",
+    "comments": "None.",
 
     "citation": "Hillis DM and Wilcox TP (2005) Molecular Phylogenetics and Evolution 34(2):299-314",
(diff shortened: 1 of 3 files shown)


gaurav:
This pull request doesn't have enough separate features to justify a separate review. I've moved all the relevant changes into #34, so I'm closing this pull request.

Issue #35, opened 2018-04-06
Create classes for some objects in Javascript


gaurav:
There are a number of functions that work on ScientificNames, Specimens, Nodes, TUnit and TUMatch objects that could be simplified if we created Javascript classes that could be easily converted into and out of their JSON representations. Since the JSON representation is a reactive model in Vue.js, we'd have to make sure that either object representations update when the JSON representation does, or we'd have to create the object from the JSON representation whenever we need it. In either case, we would have much better organized code that is easier to test.

These classes would also be a first step towards a Phyloreferencing Javascript library that could be used to load, edit, match specifiers and resolve phyloreferences against phylogenies.

gaurav:
I ended up taking the latter approach: the `./js/phyx.js` Javascript file consists of a series of wrapper classes that wrap Javascript objects representing scientific names, specimens, taxonomic units, phylogenies, phyloreferences or an entire PHYX document. The Vue.js model wraps parts of the PHYX object as needed. Modifying the contents of these objects updates the user interface through the Vue.js library. I think that covers all the requirements of this issue.

Since these changes have been incorporated into the master branch (primarily in PR #63), I will close this issue.

PR #58, opened 2018-04-30
Code cleanup with ESLint

gaurav:
This pull request incorporates ESLint so that we can enforce the [Airbnb Style Guide](https://github.com/airbnb/javascript), as well as the fixes resulting from that incorporation. 

Since this involved changing the indentation of curation-tool.js (from 4 spaces to [2 spaces](http://airbnb.io/javascript/#whitespace)), the diffs are particularly hard to read, but ESLint is now run automatically by TravisCI to ensure that they remain unchanged from here on. This PR can be merged once reviewed.


gaurav:
Thanks for the review!

> I still hate the `testWhetherXYZMatchesABCandSomethingElse` name monsters (for one, the first 11 letters are redundant with the rest of the name), but I guess it's minor and cleaning that up can be deferred to later.

That's covered by #35, and I'm working on it right now -- see 0e7049c11c364! I considered trying to do that in this pull request, but (1) I've never used classes in Javascript before, so it'll probably take me a while to get them right, and (2) having separated classes will make testing much easier, so I can start working on #4 at the same time.

> I'm also not sure I'm seeing a change to 2-space indents from 4-space. Note that you don't have to do that ,it's simply a matter of configuring ESLint to say what you indentation should be. Personally, I think 4-space indents are more suitable for JS code in terms of furthering readability, but I'll learn to deal with 2-space indentation.

True, but I wanted to stick with the Airbnb Style Guide defaults unless there was a good reason not to. It does help fit more horizontal space into a single line, which is handy since Airbnb recommends that [lines should be less than 100 characters wide](http://airbnb.io/javascript/#whitespace--max-len).

Issue #60, opened 2018-05-19
Modify the Curation Tool so it can be used over HTTPS


gaurav:
The Curation Tool is designed to work [over HTTP](http://www.phyloref.org/curation-tool/), but breaks if accessed [over HTTPS](https://www.phyloref.org/curation-tool/).

PR #61, opened 2018-05-22
Release 0.1

gaurav:
Completes release 0.1 of the Curation Tool by closing #32 and #60 and adding a change-log. When merged, this release should be tagged as "v0.1".

### DIFF
--- a/index.html
+++ b/index.html
@@ -178,7 +178,7 @@
                         <div class="input-group">
                             <input v-model.lazy="phylogeny['description']" type="text" class="form-control" placeholder="Enter a description for this phylogeny, including links">
                             <span class="input-group-btn">
-                                <button class="btn btn-default" type="button"  @click="phylogenyDescriptionBeingEdited = undefined">Done editing</button>
+                                <button class="btn btn-default" type="button" @click="phylogenyDescriptionBeingEdited = undefined">Done editing</button>
                             </span>
                         </div>
                     </template>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -666,7 +667,7 @@ <h4 class="modal-title">Editing {{selectedTUnitListContainer.type}} {{selectedTU
     ></script>
 
     <!-- Phylotree -->
-    <script src="http://d3js.org/d3.v3.min.js"></script>
+    <script src="https://d3js.org/d3.v3.min.js"></script>
```
hlapp: It may be better to just start the URL with `//`. Then the same protocol of the containing site will be used, i.e., HTTP or HTTPS.

This is based on the rationale that there is nothing won in terms of encryption by using SSL links for included content if the containing page isn't under SSL.

gaurav: I think the current best-practice is to [use HTTPS URLs everywhere](https://www.paulirish.com/2010/the-protocol-relative-url/), as this [allows browsers to use HTTP/2 when downloading from CDNs](https://jeremywagner.me/blog/stop-using-the-protocol-relative-url/#http/2) and makes sure that files downloaded from CDNs aren't changed in transit (as happened in the [2015 attack on Github](http://www.netresec.com/?page=Blog&month=2015-03&post=China%27s-Man-on-the-Side-Attack-on-GitHub)). StackOverflow switched away from protocol-relative URLs [in 2017](https://nickcraver.com/blog/2017/05/22/https-on-stack-overflow/#mistakes-protocol-relative-urls), as it was no longer a significant speed improvement even at the scale at which they operate.

Also, I like running the development version of the Curation Tool by using `file://` URL to load `index.html` locally, which breaks protocol-relative URLs!

hlapp: Alright, nothing in the way of merging then 😄

PR #63, opened 2018-05-23
Separate some controller code into classes and add a test suite

gaurav:
I've moved some controller code into a separate file (`phyx.js`) and reorganized into a series of classes (#35), mostly wrappers around reactive Vue objects representing entites in PHYX files (such as a phylogeny, a phyloreference, a specifier or a taxonomic unit). I've also written a test suite in Mocha to continuously test these classes at import. Since these classes deal with some input data slightly differently than we used to (e.g. by using 'scientificName' instead of 'binomialName' if available), I've also updated the example PHYX files so they continue to work with the Curation Tool.


test/phylorefs.js
```diff
@@ -0,0 +1,127 @@
...
+global.Vue = require('vue');
+
+// Load d3 as a global variable so it can be accessed by both phylotree.js (which
+// needs to add additional objects to it) and phyx (which needs to call it).
+global.d3 = require('d3');
+
+if (!Object.prototype.hasOwnProperty.call(global.d3, 'layout')) {
+  global.d3.layout = {};
+}
+require('../lib/phylotree.js/phylotree.js');
```
hlapp: So this doesn't need to be kept as an object?

test/phylogenies.js
```diff
@@ -0,0 +1,143 @@
...
+
+if (!Object.prototype.hasOwnProperty.call(global.d3, 'layout')) {
+  global.d3.layout = {};
+}
+require('../lib/phylotree.js/phylotree.js');
+
+const chai = require('chai');
+const phyx = require('../js/phyx');
+
+const { assert } = chai;
```
hlapp: Is there a reason to not simply state `const assert = chai;`? You're changing the type from object to function by what you wrote.

test/phylogenies.js
```diff
@@ -0,0 +1,143 @@
...
+
+describe('PhylogenyWrapper', function () {
+  describe('#constructor', function () {
+    it('should wrap a blank object', function () {
+      assert.isOk(new phyx.PhylogenyWrapper({}));
+    });
+
+    it('should be able extract all node labels in a phylogeny', function () {
```
hlapp: Are you testing the constructor here or the methods that are actually used? It seems like the latter, even if the internal extraction mechanism _currently_ all takes place in the constructor.

test/phylorefs.js
```diff
@@ -0,0 +1,127 @@
...
+
+const chai = require('chai');
+const phyx = require('../js/phyx');
+
+const { assert } = chai;
```
hlapp: Same comment as for other file.

test/scientific-names.js
```diff
@@ -0,0 +1,41 @@
+/*
+ * Test scientific name processing.
+ */
+
+/* eslint-env mocha */
+
+const chai = require('chai');
+const phyx = require('../js/phyx');
+
+const { assert } = chai;
```
hlapp: Same comment as for other file.

test/phylorefs.js
gaurav: Nope, it just adds itself to global.d3.layout! I've improved the comments to better document that.

test/phylogenies.js
gaurav: `const { assert } = chai;` is identical to `const assert = chai.assert;`, but eslint prefers object destructuring, which is why I use the former syntax. I could just use `chai.assert.isOk(...)` everywhere, but I think `assert` is pretty unambiguous.

test/phylogenies.js
gaurav: You're right! I've moved this into a '#getNodeLabels' section.

gaurav:
@hlapp Thanks for the review! I've filed an issue to add behavior-based tests to the Curation Tool test suite (#75), and I'm going to merge this assuming that I've adequately explained `const { assert } = chai;` If not, let me know and I'll fix it ASAP!

test/phylogenies.js
hlapp: I see what you mean now, I missed that. This means we require at least node 6.x, right?

gaurav: Yup, [that's right](https://node.green/#ES2015-syntax-destructuring--assignment-with-objects)! I've filed a bug so I update `package.json` to require that (phyloref/curation-tool#76).

Issue #59, opened 2018-05-08
Allow curators to sign in using ORCID's Public API


gaurav:
We need to record which curators were involved in the development of PHYX files. However, we don't need to record detailed information on exactly how each curator modified each phyloreference or which comment was edited by which curator. We will therefore implement the following sign in system:

 - The Curation Tool will allow files to be read and edited without any user being signed in.
 - It will not allow the PHYX file to be read in JSON (or, later, for the PHYX file to be downloaded to the local computer or uploaded to Github) unless a user is signed in.
 - Users will be able to sign in using the [ORCID Public API](https://members.orcid.org/api/integrate/orcid-sign-in). Once signed in, they will immediately be able to access the edited PHYX file in JSON, which will add their ORCID to the list of contributors.

Curators will only be tracked at the level of the PHYX document: there will be a single list of all ORCID users who have contributed to a particular document, but no record of who added a particular phylogeny, phyloreference or comment. Once we support uploading PHYX files directly to Github (#29), the Git update information can be used to document with a high degree of detail who changed each document and why.

hlapp:
I find this a little confusing. Can you describe with perhaps less emphasis on the technical details what a user/curator can do without being signed in, and what they can do when signed in?

gaurav:
In retrospect, this issue *is* overly complicated and also unnecessary -- authenticating with ORCID doesn't give us any additional benefits, but authenticating with Github would allow us to create and update files directly from the Curation Tool (as per #29). I've updated this issue to reflect this.

gaurav:
Klados currently records:
- A curator name (in the `curator` field in the top-level Phyx object)
- A curator e-mail address (in the `curatorEmail` field)
- A curator ORCID (in the `curatorORCID` field)

None of these fields are currently included in the Phyx specification, which instead provides a top-level `citation` field that is intended to be a citation to the ORCID file itself.

## Proposed changes
1. Remove these three individual fields and replace them with a single Citation that is described as the "Citation to this Phyx file".
2. Add support for including author e-mail addresses and ORCIDs to the Citation object.

Issue #46, opened 2018-04-14
Add '@context' to produced PHYX file


gaurav:
Currently, PHYX files have no context, and so are JSON files rather than JSON-LD files. We should add a `@context` for providing some information on how the terms have been defined in RDF. We could concatenate the entire context here, but it would probably be easier to point to a fixed URL that holds the most recent version of the `@context` we use (phyloref/curation-workflow#25).

Issue #25, opened 2018-03-27
Add phyloreference state so we can see if they are incomplete, unvalidated or validated


gaurav:
The phyloreference curation process should probably look like this:
1. Create new, blank phyloreference [state: Incomplete]
2. Add name and clade definition (see #18 for an attempt to parse the clade definition directly into specifiers).
3. Annotate note where this phyloreference is expected to resolve.
4. Add specifiers.
5. Mark phyloreference as complete [state: Unvalidated]
6. Start reasoning to test resolution. State will either change to [state: Failed Validation] or [state: Validated].

This is the basis of colour-coding phyloreferences: incomplete (clear?) and validated (green) phyloreferences are clear, while unvalidated can be further subdivided into "unvalidated but all specifiers match" (yellow) and "unvalidated but some specifiers don't match" (red).

Furthermore:
- [ ] Phyloreferences in the "draft" state that resolve as expected can automatically be marked as "final-draft"
- [ ] Phyloreferences that are marked as "published" should include a citation to their publication

hlapp:
> 5. Mark phyloreference as complete [state: Unvalidated]
> 6. Start reasoning to test resolution. State will either remain unvalidated or change to Validated.

Isn't there a difference between a phyloreference that hasn't been validated yet ("unvalidated"), and one that has been but failed to validate?

gaurav:
Remember that a phyloreference could remain in "Unvalidated" even after testing, for example if the authors did not annotate where they expected a clade definition to resolve on any phylogeny. That seems unlikely, but I'm sure it'll turn up.

gaurav:
Note that [state: Failed Validation] is necessary to the Clade Ontology test suite -- it allows us to mark phyloreferences that we don't expect to currently resolve, either because of technical limitations (such as phyloref/clade-ontology#27) or because of some ambiguity or error in the phyloreference statement itself. In the Clade Ontology, this is being tracked as phyloref/clade-ontology#31.

gaurav:
We could use the [Publication Status Ontology](https://sparontologies.github.io/pso/2014-02-01/pso.html) to give each phyloreference a publication status.
- All phyloreferences start at [pso:draft](https://sparontologies.github.io/pso/2014-02-01/pso.html#d4e640).
- Once the curator considers them complete, they hit a "Complete" button which changes its status to [pso:final-draft](https://sparontologies.github.io/pso/2014-02-01/pso.html#d4e680).
[...]

PR #66, opened 2018-06-06
Added simple phyloreference state system

gaurav:
This allows each phyloreference to be in one of the states laid out in #25. Changes in state for each phyloreference are recorded along with a timestamp, allowing the previous history of the phyloreference to be documented. In the future, this could be extended to archive previous versions of the phyloreference or to identify who changed its state (#59).

Since this pull request requires new prefixes in the PHYX file, I've also included a fix for #46 so that we include a copy of the context we need in this repository. I've incorporated [Brochu 2003](http://dx.doi.org/10.1146/annurev.earth.31.100901.141308) as an example for demonstrating phyloreference statuses.

This has been tested in phyloref/clade-ontology#29 and is ready to be merged.

### DIFF
--- a/examples/brochu_2003.json
+++ b/examples/brochu_2003.json
@@ -41,6 +41,11 @@
                     "expectedPhyloreferenceNamed": [
                         "Crocodyloidea"
                     ]
+                },
+                "Osteolaemus tetraspis": {
+                    "expectedPhyloreferenceNamed": [
+                        "Osteolaeminae"
+                    ]
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


hlapp:
> this pull request is complete, but hasn't yet been fully tested.

Is this still true?

gaurav:
Yes, still true! I've tested this in the Curation Tool by itself, but I'd like to make sure that the status information can actually be used by the Clade Ontology. I've gotten it to the point where the status information is read and retained by the Clade Ontology scripts (phyloref/clade-ontology#32), but I still need to modify JPhyloref to take it into account when testing (phyloref/jphyloref#8).

I'll submit all three pull requests for review once I have the Travis report to show that they all worked!

gaurav:
Nope, nothing controversial! Thanks for the review!

PR #73, opened 2018-07-02
Export PHYX as JSON-LD for reasoning

### LINKED ISSUE #72: Move phyloreference status code into the new controller library
We should move the phyloreference status code into the separated controller code in #63. This will allow us to write tests for it and to export phyloreference statuses in the JSON-LD output.

gaurav:
The first step in allowing the Curation Tool to carry out reasoning is to modify it to be able to produce JSON-LD that can be automatically converted into an OWL ontology that can be tested using JPhyloRef. This pull request adds a button to the Curation Tool user interface that allows the study to be downloaded as a JSON-LD file.

I'm holding off on testing the JSON-LD file produced as we can perform ad-hoc testing by converting it into RDF/XML using rdfpipe and then seeing if phyloreferences are resolved as expected using JPhyloRef, but implementing such testing will be tracked by #83.

### DIFF
--- a/index.html
+++ b/index.html
@@ -68,6 +68,10 @@
                 <div class="btn-group" role="group">
                     <button id="display-as-json-btn" type="button" class="btn btn-success" onclick="$('#display-as-json').toggle(300)">Display as JSON</button>
                 </div>
+
+                <div class="btn-group" role="group">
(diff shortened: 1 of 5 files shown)


gaurav:
I always think of the JSON-to-JSON-LD code as "controversial" because there's so many ways of do it and I'm never sure if we've gotten it right, but apart from that, nope! Thanks for the review!

PR #100, opened 2018-10-04
Fix the phylogeny spacing bug

gaurav:
A bug in the current code causes all phylogenies to share the same spacingX and spacingY variables. This PR updates the code to ensure that spacingX and spacingY can be modified on a per-phylogeny basis, and adds a test to ensure that spacingX and spacingY can be set individually for each phylogeny.

Ready for review!

### DIFF
--- a/js/curation-tool.js
+++ b/js/curation-tool.js
@@ -947,10 +947,6 @@ const vm = new Vue({
 
       let spacing = this.phylogenySpacingX[phylogenyIndex] + changeBy;
       this.phylogenySpacingX[phylogenyIndex] = spacing < MIN_SPACING_X ? MIN_SPACING_X : spacing;
-
-      if (this.phylogenySpacingX[phylogenyIndex] < MIN_SPACING_X) {
-        this.phylogenySpacingX[phylogenyIndex] = MIN_SPACING_X;
-      }
     },
 
     getPhylogenySpacingY(phylogenyIndex) {
@@ -973,10 +969,6 @@ const vm = new Vue({
 
       let spacing = this.phylogenySpacingY[phylogenyIndex] + changeBy;
       this.phylogenySpacingY[phylogenyIndex] = spacing < MIN_SPACING_Y ? MIN_SPACING_Y : spacing;
-
-      if (this.phylogenySpacingY[phylogenyIndex] < MIN_SPACING_Y) {
-        this.phylogenySpacingY[phylogenyIndex] = MIN_SPACING_Y;
-      }
     },
 
     // Methods for creating new, empty data model elements.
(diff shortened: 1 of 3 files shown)


gaurav:
I've added `changePhylogenySpacingX/Y` functions to replace manipulating `phylogenySpacingX/Y[phylogeny]` directly. I think the code looks better!

@hlapp Could I annoy you to have another look at them and let me know if you think it's clearer now?

js/curation-tool.js
```diff
@@ -892,23 +902,81 @@ const vm = new Vue({
...
+    },
+
+    changePhylogenySpacingX(phylogenyIndex, changeBy) {
+      // Change the phylogeny spacing in the X dimension by the changeBy value.
+      // We prevent the value from going below MIN_SPACING_X.
+      if (!this.hasProperty(this.phylogenySpacingX, phylogenyIndex)) {
+        Vue.set(this.phylogenySpacingX, phylogenyIndex, DEFAULT_SPACING_X);
+      }
+
+      this.phylogenySpacingX[phylogenyIndex] += changeBy;
```
hlapp: The subsequent conditional adjustment shows that this line may be setting the value of the property to a disallowed or unacceptable value. This may in reality be entirely inconsequential right now because the adjustment to be within acceptable limits follows immediately, i.e., probably at most within a few milliseconds. Note however that it's not defensive programming: if you had someone watching the object property, they'd see a flicker into disallowed territory. It's easy to avoid by just using a temporary variable:
```suggestion
      let spacing = this.phylogenySpacingX[phylogenyIndex] + changeBy;
      this.phylogenySpacingX[phylogenyIndex] = spacing < MIN_SPACING_X ? MIN_SPACING_X : spacing;
```

hlapp: It seems BTW that "Suggested change" doesn't let me indicate that this should replace several lines, so you'll have to edit it to delete the following if condition if you want to accept it 😃

js/curation-tool.js
```diff
@@ -892,23 +902,81 @@ const vm = new Vue({
...
+    },
+
+    changePhylogenySpacingY(phylogenyIndex, changeBy) {
+      // Change the phylogeny spacing in the Y dimension by the changeBy value.
+      // We prevent the value from going below MIN_SPACING_Y.
+      if (!this.hasProperty(this.phylogenySpacingY, phylogenyIndex)) {
+        Vue.set(this.phylogenySpacingY, phylogenyIndex, DEFAULT_SPACING_Y);
+      }
+
+      this.phylogenySpacingY[phylogenyIndex] += changeBy;
```
hlapp: Same as above (and would also need the following if condition to be deleted):
```suggestion
      let spacing = this.phylogenySpacingY[phylogenyIndex] + changeBy;
      this.phylogenySpacingY[phylogenyIndex] = spacing < MIN_SPACING_Y ? MIN_SPACING_Y : spacing;
```

PR #104, opened 2018-10-09
Delete taxonomic units in the Taxonomy Unit Editor panel

### LINKED ISSUE #54: Allow taxonomic units to be deleted
I seem to have forgotten to put in a delete button for taxonomic units. There should be one!

gaurav:
Updates the Taxonomic Unit Editor to include delete buttons for individual taxonomic units, so they can be deleted if necessary. Closes #54.

### DIFF
--- a/js/curation-tool.js
+++ b/js/curation-tool.js
@@ -476,12 +476,13 @@ const vm = new Vue({
       // We've set up a data-dismiss to do that, so we don't need to do
       // anything here, but the function is here in case we need it later.
     },
-    deleteTUnit(tunit) {
+    deleteTUnit(tunitToDelete) {
+      const tunit = tunitToDelete;
+
       // Delete the specified taxonomic unit from the selectedTUnitListContainer.
       this.confirm('Are you sure you want to delete this taxonomic unit? This cannot be undone!', () => {
         // Is this the selectedTUnit? If so, reset it.
-        if (this.selectedTUnit === tunit)
-            this.selectedTUnit = undefined;
+        if (this.selectedTUnit === tunit) this.selectedTUnit = undefined;
 
         // Before deleting this taxonomic unit from its list, let's wipe it in
         // the UI. Otherwise, Vue uses the values in the UI to set the values
@@ -491,9 +492,8 @@ const vm = new Vue({
         tunit.specimens = [];
 
         // Delete it from its list.
-        this.selectedTUnitListContainer.list.splice(
-          this.selectedTUnitListContainer.list.indexOf(tunit), 1
-        )
+        const list = this.selectedTUnitListContainer.list;
+        list.splice(list.indexOf(tunit), 1);
       });
     },
 
--- a/js/curation-tool.js
+++ b/js/curation-tool.js
@@ -516,10 +516,14 @@ const vm = new Vue({
 
     // Methods for building human-readable labels for model elements.
     getTaxonomicUnitLabel(tu) {
-      return new TaxonomicUnitWrapper(tu).label;
+      const label = new TaxonomicUnitWrapper(tu).label;
+      if (label === undefined) return 'Empty taxonomic unit';
+      return label;
     },
     getSpecifierLabel(specifier) {
-      return PhylorefWrapper.getSpecifierLabel(specifier);
+      const label = PhylorefWrapper.getSpecifierLabel(specifier);
+      if (label === undefined) return 'Empty specifier';
+      return label;
     },
     getPhylorefStatus(phyloref) {
       return new PhylorefWrapper(phyloref).getCurrentStatus();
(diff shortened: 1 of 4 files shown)


js/curation-tool.js
```diff
@@ -476,6 +476,26 @@ const vm = new Vue({
       // We've set up a data-dismiss to do that, so we don't need to do
       // anything here, but the function is here in case we need it later.
     },
+    deleteTUnit(tunitToDelete) {
+      const tunit = tunitToDelete;
```
hlapp: I know there's a school of thought out there that says `const` doesn't mean immutable, it only means it's like `let` except I declare not to reassign the value. Is that what you're trying to follow?

I guess I'm OK with that, though the more traditional (outside of JS) connotation of `const` is immutable.

js/phyx.js
```diff
@@ -368,8 +368,8 @@ class TaxonomicUnitWrapper {
       });
     }
 
-    // If we don't have any properties of a taxonomic unit, return undefined.
-    if (labels.length === 0) return undefined;
+    // If we don't have any properties of a taxonomic unit, return 'Empty taxonomic unit'.
+    if (labels.length === 0) return 'Empty taxonomic unit';
```
hlapp: Can you explain? This seems like a bad idea in multiple ways:
- It defines _'Empty taxonomic unit'_ by fiat in a comment as a string with very special meaning. Put differently, the comment doesn't really make sense anymore, because the part `return 'Empty taxonomic unit'` adds absolutely zero information on top of the following line of code and can just be removed, but without it the comment that remains is a half sentence left dangling.
- It hardcodes the special meaning string rather than using a constant defined elsewhere.

So how is returning this special valued string more consistent with what one should expect to be returned than `undefined`?

test/taxonomic-units.js
```diff
@@ -14,7 +14,7 @@ describe('TaxonomicUnitWrapper', function () {
     it('should wrap a blank object', function () {
       const wrapper = new phyx.TaxonomicUnitWrapper({});
       assert.exists(wrapper);
-      assert.isUndefined(wrapper.label);
+      assert.equal(wrapper.label, 'Empty taxonomic unit');
```
hlapp: Well, here we go. Now we will have to change the value in two places if we wanted to change it in the future. Two is one too many.

js/curation-tool.js
gaurav: Yup, this is how Airbnb's style guide likes to work [with complex types](https://github.com/airbnb/javascript#types--complex) -- note that the example they use has a "const" array whose values are changed! I'm fine with `const` meaning "reference doesn't change" rather than "object is immutable", but I'd be happy to try out the latter if you think it'd make the code more readable -- it looks like we'd just have to install and turn on [`eslint-plugin-const-immutable`](https://www.npmjs.com/package/eslint-plugin-const-immutable).

js/phyx.js
gaurav: You're right, and I made the same mistake in the getSpecifierLabel code. I've fixed both of them in f52356a -- the model (phyx.js) now returns undefined for (1) an "empty" taxonomic unit (i.e. one without a scientific name, specimen identifier or external link), (2) a specifier that contains only empty taxonomic units or (3) a specifier that contains no taxonomic units at all. However, the controller (curation-tool.js) will replace undefined labels with a text string (e.g. "Empty specifier") when displaying it in the view (index.html). I could have incorporated it into the view directly, but that would make the code there overly complicated. What do you think?

js/curation-tool.js
hlapp: Since you'll be the main programmer on the project, I'll leave the decision to you. If it were up to me, I'd use `const` in the meaning of immutable (rather than just not reassignable) and `let` otherwise. But I'm old school.

Note that the eslint plugin is only needed to actually enforce (in the sense of checking) immutability. Quite useful, but not strictly mandatory if you were concerned about the installation or runtime that's involved.

js/phyx.js
hlapp: It seems like how to display something is the responsibility of the view, and not the model or the controller?

js/curation-tool.js
gaurav: I think using `const` to mean "cannot be reassigned" as compared to "immutable" makes sense within the language specification -- in JavaScript, you need to explicitly use [Object.freeze()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) to make an object fully constant, and that [the specification refers only to the binding of the variable](https://mathiasbynens.be/notes/es6-const). So, I think we'll stick with `const` to mean only that the binding is immutable, not the variable itself, i.e. we'll leave [ESLint's `prefer-const` rule turned on](https://eslint.org/docs/rules/prefer-const) for now.

It also seems to be pretty ingrained within the JavaScript community (as evidenced by a quick survey of "const vs let" search results on Google and [the relevant post on StackExchange Software Engineering](https://softwareengineering.stackexchange.com/q/278652/122)). This will also make it easier for other JavaScript developers to join in on our project.

PR #105, opened 2018-10-09
Added support for jumping to the node that a phyloref is expected to/actually resolves to

### LINKED ISSUE #13: Add a "scroll to phyloreference" dropdown next to the phyloreference name
We need a dropdown because you might want to scroll to this either the expected or resolved phyloreference on any of the current phylogenies.

gaurav:
Added a button to the expected/actually resolved node display that can be clicked to jump to that node in the phylogeny. These elements are identified in the SVG display as `#expected_phyloref_node_${phylogenyIndex}` and `#pinning_node_${phylogenyIndex}` in line with existing class/variable names. Closes #13.


hlapp:
Did you want me to review this?

PR #106, opened 2018-10-09
Modified JSON-LD output to produce files without a base URI

### LINKED ISSUE #92: Replace URIs in the JSON-LD output with blank nodes
The JSON-LD export implemented in #73 uses base URIs when exporting the PHYX file for reasoning. Instead of generating these URIs, it would be better if we could create them as [blank nodes](https://en.wikipedia.org/wiki/Blank_node) instead, since they are intended to be specific to this one ontology while reasoning.

I tried to implement this when writing #73, but blank nodes were not being reasoned over correctly and so I went with using `example.org` base URIs as an expediency. This issue covers either replacing those with blank nodes or showing definitively that they cannot be used in this way.

gaurav:
Changed output JSON-LD file to have no base URI, i.e. to only have relative URIs as identifiers. These can be convert into OWL by rdfpipe or read by JPhyloRef after applying phyloref/jphyloref#26. Closes #92.

Ready for review!

### DIFF
--- a/js/phyx.js
+++ b/js/phyx.js
@@ -1454,7 +1454,7 @@ class PHYXWrapper {
 
   static get BASE_URI() {
     // Returns the default base URI for PHYX documents in JSON-LD.
-    return 'http://example.org/produced_by_curation_tool';
+    return '';
   }
 
   static getBaseURIForPhyloref(phylorefCount) {


gaurav:
Thanks! I'll merge this once phyloref/jphyloref#26 has been merged into JPhyloRef!

PR #109, opened 2018-10-17
Added individual warnings to save individual components

gaurav:
This PR adds individual warnings on phylogenies and phyloreferences that warns the user that changes made to this component have not yet been saved. Clicking the "Save as JSON" button resets these warnings, giving users a better idea of what changes have been saved and what remains to be saved. It also adds some text to the TUnit Editor to ensure that users are aware that the contents of that dialog box do not need to be saved separately.


gaurav:
Thanks, Hilmar! I've replaced "Please save it locally." with "Use 'Save as JSON' to save your changes." -- is that clearer?

hlapp:
> I've replaced "Please save it locally." with "Use 'Save as JSON' to save your changes." -- is that clearer?

Yes, absolutely.

PR #114, opened 2018-11-19
Created phyx2jsonld.js script for PHYX to JSON-LD conversion

gaurav:
This PR adds a new command-line Node.js script (`phyx2jsonld.js`) that uses the `phyx.js` library to convert an input PHYX file to a JSON-LD file that can be used for reasoning.


js/phyx2jsonld.js
```diff
@@ -0,0 +1,108 @@
...
+// phylotree.js does not export functions itself, but adds them to global.d3.layout.
+// So we set up a global.d3.layout object for them to be added to, and then we include
+// phylotree.js ourselves.
+if (!Object.prototype.hasOwnProperty.call(global.d3, 'layout')) {
+  global.d3.layout = {};
+}
+require('../lib/phylotree.js/phylotree.js');
+
+// Load phyx.js, our PHYX library.
+const phyx = require('../js/phyx');
```
hlapp: Isn't this just in the same directory? Why go one up first?

js/phyx2jsonld.js
```diff
@@ -0,0 +1,108 @@
...
+// needs to add additional objects to it) and phyx.js (which needs to call it).
+global.d3 = require('d3');
+
+// phylotree.js does not export functions itself, but adds them to global.d3.layout.
+// So we set up a global.d3.layout object for them to be added to, and then we include
+// phylotree.js ourselves.
+if (!Object.prototype.hasOwnProperty.call(global.d3, 'layout')) {
+  global.d3.layout = {};
+}
+require('../lib/phylotree.js/phylotree.js');
```
hlapp: I guess it's a little weird to see this loaded with absolute path rather than just be a part of the dependencies that get installed through `package.json` and npm. Is there a reason it is not the standard way?

js/phyx2jsonld.js
```diff
@@ -0,0 +1,108 @@
...
+const inputFilenames = args._;
+if (inputFilenames.length > 1) {
+  process.stderr.write(`phyx2jsonld.js only supports a single input PHYX file, but ${inputFilenames.length} provided: ${inputFilenames}\n`);
+  process.exit(1);
+} else if (inputFilenames.length === 0) {
+  getStdin().then(str => writeResult(convertPHYXToJSONLD(str)));
+} else {
+  fs.readFile(inputFilenames[0], 'utf8', (err, data) => {
+    if (err === null) {
+      writeResult(convertPHYXToJSONLD(data));
```
hlapp: I assume you're aware that `writeResult()` is async because it calls an async function, and there is no callback here. This will work in practice because of the place in the code that this is currently at (i.e., when this returns, the process will continue to hang on until `fs.writeFile()` terminates, but this can cause unintended effects if you add other stuff afterwards, as opposed to right now where it's at the end of the "main" thread.

js/phyx2jsonld.js
gaurav: Nope, there isn't! Before we had an npm setup, I just included the JavaScript and CSS files directly in the repository. I've opened an issue to move us to using the npm-published version of Phyloref.js (#121).

js/phyx2jsonld.js
gaurav: My bad: I figured that it'd be okay since the program never tries to interact with that file after writing it, but you're right, it's best to be sure that the file has been fully written by the time `writeResult()` returns. I've replaced `fs.writeFile()` with `fs.writeFileSync()`.

PR #116, opened 2018-11-28
Add support for multiple external specifiers

gaurav:
This PR implements multiple external specifiers by:
1. Using our current method for a single external specifier to calculate an initial clade, rather than a single pinning node.
2. Intersecting that clade with the clade formed from each subsequent external specifier. This will always result in a clade smaller than or equal to the initial clade.

Note that this method returns the set of all nodes included in the clade, not just the pinning node.

**In progress**: Still needs to be thoroughly tested, and we need to decide whether to the other kinds of phyloreferences and the Curation Tool should also return the entire clade instead of just the pinning node.

WIP


gaurav:
This code was written to work with model 1.0 ontologies, but model 2.0 ontologies don't need anything special to work with multiple specifiers correctly. Closing.

PR #137, opened 2019-01-29
Delete phyx.js and pre-Vue CLI website

gaurav:
This PR deletes files that have been moved to the https://github.com/phyloref/phyx.js repository or were necessary to the implementation and testing of our previous, non-Vue CLI design.


package.json
```diff
@@ -6,9 +6,6 @@
     "lib": "lib"
   },
   "scripts": {
-    "lint": "eslint \"js/**\" \"examples/**\" \"test/**\" --ext .js,.json",
-    "pretest": "npm run lint",
-    "test": "mocha"
```
hlapp: So now there is no testing nor linting any more? I assume there's a good justification for this, but it's not what I would have expected.

PR #144, opened 2019-05-03
Add support for editing citations

gaurav:
This PR creates a Citation component that can be used anywhere that a citation needs to be edited. In the Curation Tool, this in four places: primary phylogeny and phylogeny citations in phylogenies, and the pre-existing name and phyloref definition source in phyloreferences.

This uses the same model of citations as phyloref/clade-ontology#63, i.e. one based on that currently in Regnum and on [BibJSON](http://okfnlabs.org/bibjson/).

### DIFF
--- a/src/store/modules/citations.js
+++ b/src/store/modules/citations.js
@@ -1,0 +1,78 @@
+/*
+ * Store module for handing citations.
+ *
+ * Note that citations are located as keys on objects (for example, a phyloreference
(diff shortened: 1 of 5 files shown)


src/store/modules/citations.js
```diff
@@ -0,0 +1,296 @@
+/*
+ * Store module for handing citations.
```
hlapp:
```suggestion
 * Store module for handling citations.
```

PR #147, opened 2019-05-15
Add example phylogeny with branch lengths

### LINKED ISSUE #31: Add support for boostrap values and branch lengths in phylogenies
We used to support rendering bootstrap values and branch lengths when drawing phylogenies; however, we have to convert phylogenies back into a Newick string to add support for renaming nodes (#27). This should be doable by modifying our function `get_newick_string_from_phylotree()` (added in 21fee75dc), but none of the currently curated phylogenies have both bootstrap values and branch lengths. Once we have some, we can use those to develop this feature.

gaurav:
This PR adds an additional phylogeny to the Hillis and Wilcox example file that includes branch lengths. This demonstrates both how Phyx files containing multiple phylogenies as well as phylogenies with branch lengths are rendered. Closes #31.


gaurav:
We are no longer planning to include Hillis and Wilcox as a Klados example. Closing.

Issue #85, opened 2018-07-25
Add test suite to test browser compatibility


gaurav:
So far, I have been developing primarily in Firefox 61, which will remain our only officially supported platform for a while. The Curation Tool has been used in Google Chrome without any severe bugs. However, we should write an in-browser test suite that tests whether user interface features display and operate correctly on different browsers.

hlapp:
You might find [this article](https://philipwalton.com/articles/learning-how-to-set-up-automated-cross-browser-javascript-unit-testing/) useful. And check out Selenium if you haven't yet.

gaurav:
We should also [choose our target browsers](https://cli.vuejs.org/guide/browser-compatibility.html) in Vue CLI so we generate the right polyfills and transpline the JavaScript code correctly. We're currently using the Vue defaults, which is currently "last two versions of browsers with global usage > 1%, but not IE <= 8".

gaurav:
It looks like Vue now recommends using [Cypress](https://www.cypress.io/) for both end-to-end testing as well as component testing (see [blog post on setting this up](https://www.endpointdev.com/blog/2022/12/how-to-use-cypress-for-ui-testing/)) and Vitest for unit tests (see [browser-specific testing](https://vitest.dev/guide/browser.html)).

Cypress looks like it will meet the needs of this issue. This isn't a priority for v1.0, but I wanted to add this comment to remind myself to do this at some point.

Issue #76, opened 2018-07-13
Update package.json to indicate that we use some ES6 (i.e. Node 6.0+) features


gaurav:
The test suite uses some ES6 features, such as object destructuring, which only works in [NodeJS 6 and later](https://node.green/#ES2015-syntax-destructuring--assignment-with-objects). We should [add an `engines` line to `package.json`](https://docs.npmjs.com/files/package.json#engines) to record this.

PR #150, opened 2019-06-10
Update phyx.js to use model 2.0 ontologies

### LINKED ISSUE #11: Specifier deletion is unnecessarily complicated

### LINKED ISSUE #17: Add a field for a verbatim specifier
The specifier often contains more information than can be fully converted into a taxonomic unit. These should be stored in a verbatim specifier field.

### LINKED ISSUE #30: Make the specifier text field editable
Right now, the specifier text field is not editable, and the verbatim specifier field is not parsed for taxonomic units. The text field describes the taxonomic units referenced by the specifier.

Changing it to be editable would hugely improve phyloreference curation. There are three ways of doing that:
1. *Text field edits verbatim specifier, but does not change taxonomic units*: This would only be a marginal improvement over the current approach.
2. *Text field edits verbatim specifier, completely replacing existing taxonomic units*: Editing the specifier would cause all current taxonomic units to be deleted. The newly entered specifier would be set as the verbatim specifier, and any taxonomic units that could be extracted from the new verbatim specifier would be added to the specifier.
3. *Text field edits verbatim specifier, only replacing computer-generated taxonomic units*: If the existing taxonomic units are human-generated, we retain them; if they are computer-generated, we delete them and regenerate them from the verbatim specifier.

Option 3 requires the most work, including a redesign of the phyloreference panel to list the taxonomic units themselves so as to avoid confusion when the verbatim specifier differs from the taxonomic units. However, it's probably the best solution, and so that's the one I plan to implement.

### LINKED ISSUE #95: Specifiers should be blank boxes instead of pop up forms
Specifiers should be blank boxes instead of pop up forms, with links to options somewhere on the page to add a specimen or reference or whatever. Otherwise, you can just copy and paste from the phyloreference right into the specifier box.

### LINKED ISSUE #143: Record nomenclatural code with scientific names
When specifying a scientific name, we should provide an option for setting the nomenclatural code. We could start with the four options specified in NOMEN (see phyloref/phyx.js#5) and then extend it as needed.

This is part of the specifier redesign in #95. Updating the model to support this is covered by phyloref/phyx.js#5.

gaurav:
This PR updates the version of phyx.js being used by the Curation Tool so it produces model 2.0 ontologies that can be executed on JPhyloRef using Elk as per phyloref/jphyloref#47. It updates all three example files so they use the new Phyx format (v0.2.0).

Since mode 2.0 ontologies have a very different conception of specifiers, this PR updates how specifiers are displayed and how they may be edited as a separate Vue component (#148). This closes #17, closes #30, closes #11 and closes #95.

I'd previously planned to separate out the implementation of nomenclatural codes into a separate PR, but it didn't make sense to leave an incomplete implementation in here, so that's a part of this as well now. Closes #143.

I also noticed a minor bug in where reasoning in reset, which is also fixed here.

This should be merged after #151 has been merged.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1230,11 +1230,11 @@
       }
     },
     "@phyloref/phyx": {
-      "version": "0.1.2",
-      "resolved": "https://registry.npmjs.org/@phyloref/phyx/-/phyx-0.1.2.tgz",
-      "integrity": "sha512-/tJi1VzWoWGnm4Sxqc9YnEljLdI77u+xhAc4em/ciVs/ihWAK/h9C7ZcsN2syUMwJ9TDHVol5BB3J+92XaTEUA==",
+      "version": "0.2.0",
+      "resolved": "https://registry.npmjs.org/@phyloref/phyx/-/phyx-0.2.0.tgz",
+      "integrity": "sha512-yTCOdJ6RKeR1EsgZLEh5v9YoinqB8xMRJo6kfe5WmmUFt79T+h3NoSDl8R3tXbyRePGur6d4Ba4rdOdQkcoYRA==",
       "requires": {
-        "extend": "^3.0.2",
+        "lodash": "^4.17.11",
         "moment": "^2.23.0",
         "newick-js": "^1.1.0"
       }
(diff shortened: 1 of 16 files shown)


gaurav:
Vue automatically transplines and adds polyfills that should allow our app to be used on the last two versions of browsers that more than 1% of web users use, including IE 10/Safari 12/Firefox 66/Chrome 74/Edge 17/Opera 60. Actually testing if they work correctly is the case is covered by #85. We should definitely note that the source code is in ECMAScript 6, which is covered by #76.

Issue #166, opened 2019-09-16
Reactivate ModifiedCard to warn users about exporting the Phyx file


gaurav:
We used to have a ModifiedCard component that displayed a "This phyloreference has changed! Please remember to save it." box above any modified phyloreference (and ditto for phylogenies), but it appears to have been broken by a previous update. It would be nice to have it back so people know that they have to export their Phyx file regularly.

Related to #69.

hlapp:
Yes!!

gaurav:
This seems to be working now. I'll close this for now and reopen if it stops working again (of course, automated tests via #224 would help ensure that this problem is permanently fixed).

PR #170, opened 2019-09-18
Fix modified icons

gaurav:
This PR fixes some issues with modified icons and cards, reactivating the modified card at the top of the page (#166) and correcting some of the text we use. It also fixes a bug that prevented the "Are you sure you want to leave this page?" warning from popping up correctly and adds a modified icon next to the "Save" button on the sidebar when the phyloreference has been saved.

### DIFF
--- a/src/components/sidebar/Sidebar.vue
+++ b/src/components/sidebar/Sidebar.vue
@@ -56,7 +56,7 @@
         >
           Save
           <ModifiedIcon
-            message="This Phyx file has been modified since it was loaded! Use 'Save' to save your changes."
+            message="This Phyx file has been modified after it was loaded! Use 'Save' to save your changes."
             :compare="currentPhyx"
             :compare-to="loadedPhyx"
           />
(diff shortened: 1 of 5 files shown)


src/components/sidebar/Sidebar.vue
```diff
@@ -122,7 +127,7 @@
 
             <!-- Add a warning if this phyloreference has changed -->
             <ModifiedIcon
-              message="This phyloreference has been modified since being loaded! Use 'Save as JSON' to save your changes."
+              message="This phyloreference has been modified since being loaded! Use 'Save' to save your changes."
```
hlapp: Why _being loaded_ instead of _was loaded_ (as used above)? Though indeed, _since it was loaded_ is poor grammar, it should either be _since it has been loaded_ or _after it was loaded_. Can you fix this throughout?

gaurav: Fixed in 92a9c8d!

PR #173, opened 2019-09-27
Implement an improved syntax for expected resolution information

gaurav:
Previously, we stored expected resolution information using the additionalNodeProperties system using a custom string-based property, `expectedPhyloreferenceNamed`. This PR improves that model by storing expected resolution on each individual phyloreference and referring to a single node on a particular phylogeny. For example, the expected resolution for phyloref Gavialoidea will look like this:
```json
{
  "label": "Gavialoidea",
  "internalSpecifiers": [ ],
  "externalSpecifiers": [ ],
  "expectedResolution": {
    "#phylogeny0": {
      "nodeLabel": "Gavialis gangeticus",
      "description": "Gavialoidea resolves to a leaf node on this phylogeny"
    }
  }
}
```

This syntax allows additional information on the expectation to be stored, such as a description or whether this is intended to be the reference phylogeny (not implemented in this PR).

To maintain this data structure, Klados needs to do four things:
1. Allow the user to choose any of the labeled nodes on the phylogeny as the expected resolution node.
2. Display the expected node when rendering the phylogeny.
3. Set an identifier for the phylogeny so that it can be referred to unambiguously. By default, this is `#phylogeny` followed by the index of the phylogeny in the Phyx file, but the user interface allows this to be set by the user.
4. When a phylogeny identifier changes, all references to that phylogeny in the Phyx file should be updated as well.

As before, you don't need to provide an explicit expectation if there is a node on the phylogeny that has the same label as the phyloreference. However, the expectation syntax described here can be used to override that expected resolution.

Rewriting this system caused the `getExpectedNodeLabels()` API to change (and the method to be renamed to `getExpectedNodeLabel()`), which caused some changes in `Phylotree.vue`, which uses that information to display the expected and resolved clade on phylogenies. In addition, the `owlterms.js` module was deleted since its contents were only being used by the `resolution.js` module -- so I just merged its contents into that file instead.

This PR also includes some minor UI improvements to the display of internal node labels in `Phylotree.vue`, to the "Internal specifier"/"External specifier" header sizes in `PhylorefView.vue` and to the list of phylogenies in the sidebar.


gaurav:
Good point! I've filed that as https://github.com/phyloref/curation-tool/issues/176.

PR #180, opened 2019-12-31
Updated JPhyloRef submission URL to ggvaidya.com

gaurav:
Since our backend server hosted at UF is temporarily down, this PR changes that to my server at ggvaidya.com for now. We'll revert it to UF once it's working again.

Should be merged after #178 and #179.

### DIFF
--- a/public/examples/brochu_2003.json
+++ b/public/examples/brochu_2003.json
@@ -77,6 +77,7 @@
             ]
         },
         {
+            "@id": "#crocodylia",
             "label": "Crocodylia",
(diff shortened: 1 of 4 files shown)


gaurav:
> This seems to include quite a bit more than the title suggests, perhaps because it is against master whereas the branch it is coming from is not from the current HEAD of master?

Yup, that's right! I had to rebase those changes into this PR because that was the only way I could test them -- without changing the JPhyloRef URL, the resolution code doesn't work at all. However, it would have been cleaner to have one PR at the end whose only purpose is to combine those changes, such as https://github.com/phyloref/open-tree-resolver/pull/18. I'll do that in the future!

> Regarding the specific change in the title, please refer to my comment for the corresponding OTR change.

I don't think I'll be able to get this to work by tomorrow except by closing #183, so I'll go ahead and merge this in for now. Once we resolve #183, this shouldn't be an issue any longer.

PR #185, opened 2020-01-03
Minor UI fixes

gaurav:
This PR fixes some minor UI issues on Klados:
- There is an unexpected gray bar between phylorefs on the sidebar. This PR partially fixes this issue, eliminating the bars but without black lines to clearly separate phylorefs. 
- The primary reference phylogeny has not been clearly distinguished from the other reference phylogeny, so I've temporarily hidden it.
- One of the example files cannot be read, since it is still in the old JSON format. I've temporarily hidden it.
- The "Add specifier" button in the sidebar is non-functional and not very useful in the current design, so I've removed it entirely.

### DIFF
--- a/src/components/phylogeny/PhylogenyView.vue
+++ b/src/components/phylogeny/PhylogenyView.vue
@@ -69,12 +69,12 @@
             </div>
           </div>
 
-          <!-- Primary reference phylogeny -->
+          <!-- Primary reference phylogeny
           <Citation
             label="Primary reference phylogeny"
             :object="selectedPhylogeny"
             citation-key="primaryPhylogenyCitation"
-          />
+          /> -->
 
           <!-- Other reference phylogeny -->
           <Citation
--- a/src/components/sidebar/Sidebar.vue
+++ b/src/components/sidebar/Sidebar.vue
@@ -104,7 +104,7 @@
           Phyloreferences
         </a>
         <a
-          class="list-group-item list-group-item-action"
+          class="border-dark list-group-item list-group-item-action"
           href="javascript: void(0)"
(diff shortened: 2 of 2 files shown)


src/components/phylogeny/PhylogenyView.vue
```diff
@@ -69,12 +69,12 @@
...
           </div>
 
-          <!-- Primary reference phylogeny -->
+          <!-- Primary reference phylogeny
           <Citation
             label="Primary reference phylogeny"
             :object="selectedPhylogeny"
             citation-key="primaryPhylogenyCitation"
-          />
+          /> -->
```
hlapp: Given the description on the PR, am I right this appears to be "confusing"? If so, that fact isn't evident from the code, so I'd suggest adding a comment as to why we're commenting it out for now. Meaning this is a temporary measure, and how do we intend to fix the confusing nature; perhaps refer to an issue to be created on the tracker. Or if this is not temporary, why not remove it altogether?

gaurav:
> But there's some evidence of hastiness here, and if time allows please clean this up.

Indeed. I'll clean this up next week and resubmit for review!

> Aside from the inline comment, I'm not sure anyone will remember in the future what the purpose of the by-0 and my-0 classes is that you added – it's hard to imagine less descriptive class names.

Fair. Those are standard Bootcamp class names, so anybody familiar with that *should* know what they mean. Now that I have more time, I'll try to figure out why the previously used CSS classes aren't working correctly, and whether I can correct those by changing some of the other classes being used. Otherwise, I'll add some comments so it's clearer for future maintainers!

gaurav:
Superceded by https://github.com/phyloref/klados/pull/306, closing.

PR #188, opened 2020-01-29
Create Github Action for deploying to Github Pages

gaurav:
I've based this on an existing workflow in the Github Marketplace. It should be triggered every time we publish a new release. Once triggered, it checks out that release, builds the HTML and JavaScript using Vue, and deploys it to the gh-pages branch using an SSH deploy key stored as a secret.

Since we no longer need to publish a static version of the page at `docs/`, this PR also deletes all those files. In the future, the static version will be automatically generated by GitHub Actions and stored only on the `gh-pages` branch.

See https://github.com/phyloref/curation-tool/commit/883dc2cc75d66bda82dab0b73752e5ee1f3fcba5 for an example commit generated by this workflow.


gaurav:
@hlapp Oops, I forgot to document that: we previously stored a static version of the website in the `docs/` directory. Since we no longer need that, this PR deletes the thirteen files in the `docs/` directory. From here on, we'll only generate the static content using GitHub Actions, and store it only in the `gh-pages` branch for deployment via GitHub Pages.

PR #198, opened 2020-07-14
Allow nodes to be renamed on phylogenies

gaurav:
This PR allows users to rename nodes directly on the phylogeny by adding a custom menu item to the menu that appears when clicking on a node. To make it easier to look up the reference phylogeny, it also adds the phylogeny citation to the Phyloref display.

It also fixes a minor bug (the phylogeny display now updates when the underlying Newick string changes), updates package.json so that the package name is "klados" and not "curation-tool", and upgrades all NPM packages.


src/components/phylogeny/Phylotree.vue
```diff
@@ -140,6 +140,43 @@ export default {
...
+
+              // Export the entire phylogeny as a Newick string, and store that
+              // in the phylogeny object.
+              this.$store.commit('setPhylogenyProps', {
+                phylogeny: this.phylogeny,
+                newick: tree.get_newick((n) => {
+                  // There appears to be a bug in the version of Phylotree.js we
+                  // use in which leaf nodes are duplicated if we just return n.name
+                  // here. So we return undefined for leaf nodes and n.name for
+                  // everything else.
```
hlapp, line 171: Is there a bug report filed with Phylotree.js? If yes, can this be quoted here. If not, shouldn't there be? Or if we can address this upgrading to a newer version, shouldn't we do that?

gaurav: I tried upgraded Phylotree.js in 530b379, but the bug is still present. Given how complicated our usage of Phylotree.js is, it might take a while to come up with a minimal example that demonstrates this bug, so I've filed it as #200. I'll get around to it eventually!

PR #203, opened 2020-08-28
Add support for higher-taxon match

gaurav:
The plan is to record that, for a particular phylogeny, a certain node should be considered as including a particular taxonomic unit. We can model this in CDAO by assigning multiple taxonomic units to the same node. This PR will add user interface elements to Klados to allow the curator the change which taxonomic units as assigned to which (labeled) node.

WIP

### DIFF
--- a/src/components/phyloref/PhylorefView.vue
+++ b/src/components/phyloref/PhylorefView.vue
@@ -387,32 +387,71 @@
               />
             </div>
 
-            <!-- Display which taxa include other taxa -->
-            <!-- Expected resolution information -->
+            <!-- Display and edit phylogenies that include other taxa -->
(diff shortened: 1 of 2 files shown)


gaurav:
By the time I stopped worked on this PR in August 2020, it was a very early version of #228 that allowed users to directly modify the JSON object associated with each labeled node on a phylogeny, which was always intended to be replaced by a better UI in the future. I'm developing that in PR https://github.com/phyloref/klados/pull/307 now. I've copied over the code for accessing the additionalNodeProperties to PR https://github.com/phyloref/klados/pull/307 in case that turns out to be useful; otherwise, I think it makes sense to close this PR.

PR #205, opened 2020-09-09
Upgraded NPM packages using `npm upgrade`

gaurav:
This PR upgrades all NPM packages using npm upgrade.


hlapp (approved):
Github won't let me review this ("diff too large, please use a local Git client to view these changes"). But I also don't think there can be much or anything here that would warrant a review, so I'm signing off sight unseen.

PR #229, opened 2022-01-12
Phyx and JSON-LD downloads are now named usefully

### LINKED ISSUE #163: Improve name of downloaded files
Currently, downloaded files are presented to the browser as `download.json` (Phyx file) or `download.nq` (OWL export). We can improve this by constructing a name from either the filename of the input JSON file or from the name of the phyloreferences in this file (e.g. `Eusuchia_Crocodylia_and_20_others.json`).

gaurav:
This PR names downloads (whether JSON or JSON-LD) from Klados based on the labels of the included phyloreferences.

Closes #163.

### DIFF
--- a/src/components/sidebar/Sidebar.vue
+++ b/src/components/sidebar/Sidebar.vue
@@ -242,6 +242,24 @@ export default {
     };
   },
   computed: {
+    downloadFilenameForPhyx() {
+      // Return a filename to be used to name downloads of this Phyx document.
+      const DEFAULT_DOWNLOAD_FILENAME = 'download';
+
+      if (!this.currentPhyx || !this.phylorefs) {
+        return DEFAULT_DOWNLOAD_FILENAME;
+      }
+
+      // Determine all phyloref labels in this document.
+      const phylorefLabels = this.phylorefs.flatMap((p, index) => (has(p, 'label') ? [p.label.replaceAll(/\W/g, '_')] : [`Phyloref_${index+1}`]));
+
+      if (phylorefLabels.length === 0) return DEFAULT_DOWNLOAD_FILENAME;
+
+      if (phylorefLabels.length === 1) return phylorefLabels[0];
+      if (phylorefLabels.length === 2) return `${phylorefLabels[0]}_and_${phylorefLabels[1]}`;
+      if (phylorefLabels.length === 3) return `${phylorefLabels[0]}_${phylorefLabels[1]}_and_${phylorefLabels[2]}`;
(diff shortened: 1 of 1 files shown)


src/components/sidebar/Sidebar.vue
```diff
@@ -372,8 +399,8 @@ export default {
       const content = [JSON.stringify(this.$store.state.phyx.currentPhyx, undefined, 4)];
 
       // Save to local hard drive.
-      const jsonFile = new File(content, 'download.json', { type: 'application/json;charset=utf-8' });
-      saveAs(jsonFile);
+      const jsonFile = new File(content, `${this.downloadFilenameForPhyx}.json`, { type: 'application/json;charset=utf-8' });
+      saveAs(jsonFile, `${this.downloadFilenameForPhyx}.json`);
 
       // saveAs(jsonFile) doesn't report on whether the user acceped the download
```
hlapp, line 405: Note the (very small) typo (_acceped_).

hlapp (approved):
Looks good. I suppose we're ignoring the possibility that the file name could become quite long (meaning, there's no check triggering truncation if a certain length is exceeded). But supposedly that should be fine on any modern OS (including, presumably, Windows?).

Note also the very minor typo (already present, not introduced in this change). I'd suggest leaving it there (since it's unrelated to what it being changed here) but the pedant in me can't refrain from marking it.

src/components/sidebar/Sidebar.vue
gaurav: Fixed in b6c695be34a6.

gaurav:
> Looks good. I suppose we're ignoring the possibility that the file name could become quite long (meaning, there's no check triggering truncation if a certain length is exceeded). But supposedly that should be fine on any modern OS (including, presumably, Windows?).

Alas, it looks like Windows by default limits (file path + file name) to 260 characters ([as per their docs](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation)). Theoretically, that shouldn't be a direct issue for us -- we only suggest the download filename in the [Content-Disposition HTTP field](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition), and it looks like the browser will then [truncate the filename](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) to whatever the local filesystem supports.

Overlong filenames are unlikely to be a problem, since I don't think clade definitions are likely to more than 100 characters long. I propose we wait until someone runs into this problem rather than trying to come up with a made-up example ourselves, but if you think we should open an issue for that, please go ahead and do that.

PR #230, opened 2022-01-12
Export as N-quads

### LINKED ISSUE #65: Replace a text-based expectation on phylogeny with an reference-based expectation as per Model 2.0
Currently, we use a term called `expected_phyloreference_named` to store the label of the phyloreference that a particular node is expected to match. However, we should instead create an `expected_phyloreference` property that allows us to link to the phyloreference by identifier rather than by label.

### LINKED ISSUE #204: Hillis and Wilcox example file doesn't work
This is probably because it uses the old syntax. Shouldn't take too long to fix.

### LINKED ISSUE #222: Add support for RDF output in N-Quads
Clicking "Export as Ontology" will currently export the ontology in JSON-LD, which can be a complicated format for users to understand or use. Instead, we should change this to export the ontology in N-Quads. This functionality is already [available in phyx.js](https://www.phyloref.org/phyx.js/class/src/wrappers/PhyxWrapper.js~PhyxWrapper.html#instance-method-toRDF), so this should just be the case of exposing this functionality.

Doing this will immediately close #159, since an OWL ontology in N-Quads can always be opened by Protege.

gaurav:
This PR upgrades `@phyloref/phyx.js` to the latest version, allowing us to use reference-based (model 2.0) expectations (#65) and to export the ontology in n-Quads (#222). A few phyx.js methods have been renamed or replaced since this code was written; this PR updates them so that phyloref resolution works again.

See #231 and #232 for some follow-up changes that will be needed after this PR.

Closes #65, closes #204, closes #222.

Should be merged after PR #229.


src/components/sidebar/Sidebar.vue
```diff
@@ -427,6 +427,20 @@ export default {
...
     },
 
+    downloadAsNQuads() {
+      // Exports the PHYX file as an OWL/N-Quads file, which can be opened in
+      // Protege or converted into other RDF formats.
+      const wrapped = new PhyxWrapper(this.$store.state.phyx.currentPhyx);
+
+      // TODO: we need a baseIRI here because of https://github.com/phyloref/phyx.js/issues/113
+      // Once that is fixed in phyx.js, we can remove it here.
+      wrapped.toRDF('https://example.org/phyx#').then((content) => {
```
hlapp, line 437: For aesthetic reasons, wouldn't something more personalized be preferable, such as `https://example.phyloref.org/phyx#`?

hlapp (approved):
👍🏻 (though see minor suggestion)

src/components/sidebar/Sidebar.vue
gaurav: Sure! I always prefer `example.org` since it's widely known to be non-resolvable, but the actual value doesn't really matter. I've changed it to the IRI you suggest in cd0e4e0.

PR #233, opened 2022-01-25
Add Open Tree of Life resolution

### LINKED ISSUE #40: Download relevant subtree from Open Tree of Life
We can very easily integrate the Open Tree of Life into the Curation Tool:
1. Make a list of all scientific names in the specifiers in the current PHYX file.
2. Filter out those scientific names are present in the Open Tree of Life (implementing #15 for the Open Tree Taxonomy would help).
3. Extract the Newick string for the subtree that includes all these scientific names from the Open Tree of Life (see [working example](https://beta.observablehq.com/@gaurav/retrieving-an-open-tree-mrca-with-phylotree-js)).
4. Add this Newick string as an additional phylogeny.

This would, if successful, allow us to very, very quickly allow us to visualize phyloreferences against the synthetic Open Tree of Life, and later to resolve them.

### LINKED ISSUE #133: Add support for indicating a primary or secondary reference phylogeny out of all the phylogenies in the Phyx file
As per phyloref/phyloref-ontology#14 and phyloref/phyloref-ontology#16, we can use [`prov:wasDerivedFrom`](https://www.w3.org/TR/prov-o/#wasDerivedFrom) to relate a phyloreference to the PHYX file from which it was generated and/or the original publication from which we curated the phyloreference. We can also use [`has evidence`](http://purl.obolibrary.org/obo/RO_0002558) to link a phyloreference to its reference phylogeny.

In terms of the PHYX file, it's probably fine to keep its current structure. However, these relationships should be generated by `phyx2ontology` when generating the Clade Ontology.

gaurav:
This PR adds a "Add Open Tree of Life phylogeny" button to the sidebar, which adds the induced subtree from the Open Tree of Life that includes all named taxa used as specifiers across all phyloreferences. Taxa that don't have OTT IDs are ignored, while OTT IDs that not included in the synthetic are ignored. Closes #40.

This appears as a button in the bottom left corner of the screen labeled "Add Open Tree of Life phylogeny":
Screen Shot 2022-03-29 at 2 48 44 AM

Clicking this button doesn't have any obvious signs that anything is happening (I'm not sure where they would go), but once the API calls have completed, a new phylogeny labeled "Open Tree of Life phylogeny" has been added to the Phyx file:
Screen Shot 2022-03-29 at 2 49 02 AM

This phylogeny is the induced subtree of all phylogeny nodes that have Open Tree of Life Taxonomy IDs:
Screen Shot 2022-03-29 at 2 49 33 AM

(Please ignore the "Primary Reference phylogeny" field; once we close #133, only the "Reference phylogeny" field will remain, and it'll be renamed to something like "Citation").

With the changes I made in 26f836ed73e8b, this now includes a citation to the Open Tree of Life synthetic phylogeny, as suggested at https://tree.opentreeoflife.org/about/open-tree-of-life -- I've included both the synthetic tree and taxonomy version numbers as reported by the Open Tree API, but only cited the synthetic tree, since we only use the taxonomy to look up nodes on the synthetic tree.
Screen Shot 2022-03-30 at 2 01 09 AM

### DIFF
--- a/src/store/modules/phyx.js
+++ b/src/store/modules/phyx.js
@@ -118,6 +118,43 @@ export default {
       // Create a new, empty phylogeny from the Open Tree of Life using the /induced_subtree endpoint
       // at https://github.com/OpenTreeOfLife/germinator/wiki/Synthetic-tree-API-v3#induced_subtree
 
+      function createOTTPhylogenyWithCitation(phylogeny) {
+        // Create an OTT phylogeny after querying for the synthetic tree ID.
+        jQuery.ajax({
+          type: 'POST',
+          url: 'https://api.opentreeoflife.org/v3/tree_of_life/about',
+          dataType: 'json',
+          error: err => console.log('Could not retrieve Open Tree of Life /about information: ', err),
+          success: (data) => {
+            // Citation as per https://tree.opentreeoflife.org/about/open-tree-of-life, retrieved March 30, 2022.
+            const citation = {
+              type: 'misc',
+              authors: [
+                {
+                  name: 'OpenTree et al',
+                },
+              ],
+              year: new Date().getFullYear(),
+              title: `Open Tree of Life synthetic tree, release ${data['synth_id'] || 'Unknown synthetic tree version'}`,
+              link: [
+                {
+                  url: 'https://doi.org/10.5281/zenodo.3937741'
+                },
+              ],
+            };
+
+            // Add the citation to the provided phylogeny.
+            const citedPhylogeny = {
+              phylogenyCitation: citation,
+              ...phylogeny,
+            };
+
+            // Now create this phylogeny via Vue.
+            commit('createPhylogeny', { phylogeny: citedPhylogeny });
+          },
+        });
+      }
+
       // Step 1. Get a list of all taxon names used across all phyloreferences.
       const taxonConceptNames = (state.currentPhyx.phylorefs || [])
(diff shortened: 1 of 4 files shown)


gaurav:
TODO:
- [ ] Remove internal nodes that only have a single child, as these just clutter up the user interface for no reason.

src/store/modules/phyx.js
```diff
@@ -108,4 +113,165 @@ export default {
...
+  actions: {
+    createPhylogenyFromOpenTree({ commit, state }) {
+      // Create a new, empty phylogeny from the Open Tree of Life using the /induced_subtree endpoint
+      // at https://github.com/OpenTreeOfLife/germinator/wiki/Synthetic-tree-API-v3#induced_subtree
+
+      function createOTTPhylogenyWithCitation(phylogeny) {
+        // Create an OTT phylogeny after querying for the synthetic tree ID.
+        jQuery.ajax({
+          type: 'POST',
+          url: 'https://api.opentreeoflife.org/v3/tree_of_life/about',
```
hlapp, line 125: Shouldn't we make the API URL path prefix a constant, so it's not repeated in multiple places?

hlapp (approved):
Looks good. See one inline suggestion.

src/store/modules/phyx.js
gaurav: Fixed in f8fa3f8b70e2b36973850a6cdc7daa977d4f53a7.

PR #234, opened 2022-01-25
Add deletion button for citations

### LINKED ISSUE #167: Add delete button for citations
Currently, once you start editing a citation, there is no way to completely delete it -- you can only set it to blank. We should add a delete button so phylorefs/phylogenies without a citation can be described correctly.

gaurav:
This PR adds a delete button to citations, while also tweaking the display of citations. It also fixes a bug in PhylorefView, where we referred to `definitionSource` as `obo:IAO_0000119`, which was its pre-publication field name. Closes #167.

Citation filled in:
Screen Shot 2022-04-26 at 9 46 00 PM

Citation expanded:
Screen Shot 2022-04-26 at 9 46 25 PM

Pressing the red 'X' at the far right displays a confirmation dialog:
Screen Shot 2022-03-29 at 1 29 05 AM

If the user selects "OK", the citation is deleted:
Screen Shot 2022-04-26 at 9 47 43 PM

The "Add citation" text is underlined if the user hovers over it:
Screen Shot 2022-04-26 at 9 48 28 PM

### DIFF
--- a/src/components/citations/Citation.vue
+++ b/src/components/citations/Citation.vue
@@ -569,10 +569,13 @@ export default {
   methods: {
     deleteCitation(index) {
       // Remove the citation at a particular index from the input citations.
-      if (Array.isArray(this.object[this.citationKey])) {
-        this.object[this.citationKey].splice(index, 1);
-      } else {
-        Vue.delete(this.object, this.citationKey);
+      // eslint-disable-next-line no-restricted-globals
+      if (confirm('Are you sure you wish to delete this citation?')) {
+        if (Array.isArray(this.object[this.citationKey])) {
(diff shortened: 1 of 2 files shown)


hlapp (commented):
So instead of &times; can't we use a trash bin icon (should be available within font-awesome, if we're using that?)? The cross is just often used for closing something, or making something  disappear from view. A trash bin would be unequivocal.

gaurav:
Updated with the trash icon from Bootstrap -- the full list of icons available to us are at https://bootstrap-vue.org/docs/icons#icons-1. I've also updated the screenshots with the new trash icon.

hlapp (approved):
Looks good!

Issue #232, opened 2022-01-25
Add additional example files


gaurav:
In PR #230, I removed the following example files:
* fisher_et_al_2007.json (from https://doi.org/10.1639/0007-2745%282007%29110%5B46%3APOTCWA%5D2.0.CO%3B2)
  - Includes an apomorphy-based definition
* hillis_and_wilcox_2005.json (from https://doi.org/10.1016/j.ympev.2004.10.007)
  - Includes specimen-based definitions
* Most of the phyloreferences from Brochu 2003, replacing it with the minimal version used in the phyx.js tests.

This leaves us with a single small example file, Brochu 2003. This issue tracks us adding additional example files (probably from the Clade Ontology) to Klados. Ideally, they should demonstrate specimen identifiers (like Fisher et al did) or apomorphy-based phyloreferences. Using files from the Clade Ontology will almost certainly be easier than attempting to convert these v0.2.0 files to v1.0.0 Phyloref files.

- [ ] Also check whether phyloreferences with specimens and external references as specifiers can be exported correctly as CSV.

Could be part of the tutorial (#227).

gaurav:
Instead of recreating the old Brochu 2003 (which contained 17 clade definitions and 5 alternate definitions), we could include the 25 clade definitions from [Sangster et al, 2022](https://www.sciencedirect.com/science/article/pii/S2053716622000238), which are open access (under the CC-BY-NC-ND license). Unlike Brochu 2003, however, they don't include their own phylogeny down to the species level showing where they expect these clades to resolve, so we would need to include them separately from [Prum et al. (2015)](https://doi.org/10.1038/nature15697) and [Kuhl et al. (2021)](https://doi.org/10.1093/molbev/msaa191).

hlapp:
Who would curate those definitions? And also, would it be wise to have different example datasets for different software products (if there aren't good reasons as to why one dataset would be much more suitable for one product than another)?

gaurav:
> Who would curate those definitions?

We could use already-curated definitions from the Clade Ontology, as long as we find examples that really exercise Klados. By exercising Klados, I mean:
- Having a variety of publication sources
- Having a variety of phylogenies
- Having a variety of specifier types: concept IRIs, apomorphies, specimens and taxa in a number of nomenclatural codes.

I think the phylorefs we've curated into JSON for the Clade Ontology has examples of all of these except for a variety of publication sources.

[...]

PR #238, opened 2022-02-16
Export Phyx view as CSV

### LINKED ISSUE #225: Add ability to export the Phyx View as a CSV
The Phyx View includes a useful summary of the state of the Phyx file: a list of all the phyloreferences in the file, with information on whether or not they resolved as expected on the reference phylogeny. It might be useful to be able to export this information as a CSV file.

gaurav:
This PR adds an "Export as CSV" button to the Phyloref summary in the Phyx view so that that table can be exported as a CSV file (see attached screenshot). The generated CSV file expands some of the fields in the visible table (see example: 
[`Alligatoridae_Alligatorinae_and_4_others.csv`](https://github.com/phyloref/klados/files/8567908/Alligatoridae_Alligatorinae_and_4_others.csv) from the screenshot below). To implement this, I moved the download filename generation code into the store so that it could be used in multiple places in this application.

If the phyloreference resolves to an unlabelled node, this will appear in the exported CSV as "(unlabelled)" (see [example CSV](https://github.com/phyloref/klados/files/8567946/Alligatoridae_Alligatorinae_and_4_others-2.csv)).

This PR also includes an unrelated fix that I noticed (`getDefaultNomenCodeURI()` referred to `NAME_IN_UNKNOWN_CODE` instead of `UNKNOWN_CODE` as the constant is currently named).

Closes #225.

Screenshot of the new button:
Screen Shot 2022-04-26 at 9 22 47 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "@phyloref/phyx": "^1.0.1",
         "bootstrap": "^4.5.0",
         "bootstrap-vue": "^2.15.0",
+        "csv-stringify": "^6.0.5",
         "filesaver.js-npm": "^1.0.1",
         "jquery": "^3.5.1",
         "lodash": "^4.17.19",
@@ -6380,6 +6381,11 @@
       "dependencies": {
         "cssom": "0.3.x"
       }
+    },
+    "node_modules/csv-stringify": {
+      "version": "6.0.5",
+      "resolved": "https://registry.npmjs.org/csv-stringify/-/csv-stringify-6.0.5.tgz",
+      "integrity": "sha512-7xpV3uweJCFF/Ssn56l3xsR/k2r3UqszwjEhej9qEn2cCPzyK1WyHCgoUVzBA792x8HbwonNX7CU9XM2K5s5yw=="
     },
     "node_modules/current-script-polyfill": {
       "version": "1.0.0",
@@ -23983,6 +23989,11 @@
       "requires": {
         "cssom": "0.3.x"
       }
+    },
+    "csv-stringify": {
+      "version": "6.0.5",
+      "resolved": "https://registry.npmjs.org/csv-stringify/-/csv-stringify-6.0.5.tgz",
+      "integrity": "sha512-7xpV3uweJCFF/Ssn56l3xsR/k2r3UqszwjEhej9qEn2cCPzyK1WyHCgoUVzBA792x8HbwonNX7CU9XM2K5s5yw=="
     },
     "current-script-polyfill": {
       "version": "1.0.0",
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "@phyloref/phyx": "^1.0.1",
     "bootstrap": "^4.5.0",
     "bootstrap-vue": "^2.15.0",
+    "csv-stringify": "^6.0.5",
     "filesaver.js-npm": "^1.0.1",
     "jquery": "^3.5.1",
     "lodash": "^4.17.19",
--- a/public/examples/brochu_2003.json
+++ b/public/examples/brochu_2003.json
@@ -68,6 +68,7 @@
   ],
   "phylorefs": [
     {
+      "@id": "#Alligatoridae",
       "label": "Alligatoridae",
       "scientificNameAuthorship": {
         "bibliographicCitation": "(Cuvier 1807)"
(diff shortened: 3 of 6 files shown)


hlapp (commented):
Nothing that looks obviously questionable here, but I don't see a screenshot of how this would appear, and I don't see an example csv that this would produce. So kind of hard to confidently give 👍🏻.

gaurav:
> Nothing that looks obviously questionable here, but I don't see a screenshot of how this would appear, and I don't see an example csv that this would produce. So kind of hard to confidently give 👍🏻.

Good catch! I've updated the PR description with a screenshot and a link to two example CSVs.

hlapp (changes requested):
UI appearance looks fine. As for the CSV, my one gripe with it is that the first column pretends that phyloreferences are identified by label, which of course they aren't. My suggestion would be to have two columns for Phyloreference ID and label, respectively, or if only one column at the very least be clear in the column label that the value is a label.

gaurav:
Fixed in d3f7582277f9. The file now includes a "Phyloreference ID" field -- if an `@id` is set, it will be used; otherwise, each phyloreference is referred to as `#phyloref0`, `#phyloref1`, and so on. See example CSV output in [Alligatoridae_Alligatorinae_and_4_others.csv](https://github.com/phyloref/klados/files/8905597/Alligatoridae_Alligatorinae_and_4_others.csv).

hlapp (approved):
Looks good with the ID present now.

Two remaining quibbles:
1. We're not including the type of specifier, so are leaving it up to the reader of the CSV to either assume it's a taxon name, or to apply some guessing logic. However, we do know the type, don't we? Same for expected and actual. Or maybe all we're saying is that these are node labels?
2. It seems that the number of internal and external specifier columns is not fixed, but dynamic depending on however many the phyloreference with the most such specifiers happens to use. I suppose that's unavoidable when flattening out to CSV. But wouldn't it be better to have the columns whose number is dynamic appear last in the column order, rather than in the middle?

I think both of these are minor quibbles, and rather than holding up this PR should probably rather be turned into issues, which is why I'm approving.

gaurav:
> We're not including the type of specifier, so are leaving it up to the reader of the CSV to either assume it's a taxon name, or to apply some guessing logic. However, we do know the type, don't we? Same for expected and actual. Or maybe all we're saying is that these are node labels?

Ah, good catch! That's a bug on my part -- I wrote `TaxonConceptWrapper` instead of `TaxonomicUnitWrapper`, which has the logic for displaying specimens (with the phrase `Specimen [occurrence ID]`) or an external reference (in the format `<url>`). Fixed in 41dd4cc. I can't really test this since our only current test file only uses taxon names, but I've made a note on #232 to check the CSV export once we have more complicated phyloreferences in here.

> It seems that the number of internal and external specifier columns is not fixed, but dynamic depending on however many the phyloreference with the most such specifiers happens to use. I suppose that's unavoidable when flattening out to CSV. But wouldn't it be better to have the columns whose number is dynamic appear last in the column order, rather than in the middle?

In theory, the phylogeny columns are also dynamic, as you are allowed to have multiple phylogenies in a single Phyx file. If there were multiple phylogenies in [Alligatoridae_Alligatorinae_and_4_others.csv](https://github.com/phyloref/klados/files/8905597/Alligatoridae_Alligatorinae_and_4_others.csv), each would be listed after all the other columns with the columns `[phylogeny label] expected` and `[phylogeny label] actual`. So even if I were to move them to before the specifier columns, they would still require some processing to be interpreted correctly.

I'll go ahead and merge this PR, but please do open an issues if you think we should output the phylogeny resolutions in CSV in another format.

hlapp:
> In theory, the phylogeny columns are also dynamic, as you are allowed to have multiple phylogenies in a single Phyx file.

Good point. So I think the main thing left is to document the CSV format, either in plain text, or for example in LinkML, so that downstream consumers know what to expect? Should that become an issue in the tracker so we don't forget?

gaurav:
> Good point. So I think the main thing left is to document the CSV format, either in plain text, or for example in LinkML, so that downstream consumers know what to expect? Should that become an issue in the tracker so we don't forget?

Sounds good! I'm not sure if LinkML supports dynamic column names in this way, but we should definitely document this somewhere. Filed as #257.

PR #256, opened 2022-08-21
Lint code and upgrade NPM packages

gaurav:
This PR carries out two automated maintenance tasks:
- Linting the source code by running `npm run lint --fix`
- Upgrading all the NPM packages using `npm upgrade`


hlapp (approved):
Looks like if I were to disapprove of anything I'd be disapproving with the style guide. 👍

PR #260, opened 2022-09-07
Store curation information and default nomenclatural code in cookies

gaurav:
While writing the tutorial, I realized that users currently need to re-enter their name, e-mail address and ORCID every time they create a new Phyx file. This PR changes that so that those four pieces of information -- as well as the default nomenclatural code -- are stored in the users' browser in cookies for 30 days.

Should be merged after PR #259.

### DIFF
--- a/src/components/phyx/PhyxView.vue
+++ b/src/components/phyx/PhyxView.vue
@@ -73,6 +73,20 @@
               </select>
             </div>
           </div>
+
+          <!-- Checkbox for permission to save this information to browser -->
+          <div class="form-group row">
+            <div class="col-md-2">&nbsp;</div>
+            <div class="col-md-10">
+              <input type="checkbox"
+                     :checked="cookieCheckbox"
+                     @click="$store.commit('toggleCookieAllowed')"
+              />
+              Click this checkbox if you would like Klados to save your curator information and
+              default nomenclatural code as a browser cookie for the next month, or uncheck it to
+              delete all Klados cookies from your browsers.
+            </div>
+          </div>
         </form>
       </div>
     </div>
@@ -279,6 +293,9 @@ export default {
     phyxCuratorORCID: {
       get() { return this.phyx.curatorORCID; },
       set(orcid) { this.$store.commit('setCurator', { orcid }); },
+    },
+    cookieCheckbox() {
+      return this.$store.getters.isCookieAllowed;
     },
     ...mapState({
       phyx: state => state.phyx.currentPhyx,
--- a/src/components/phyx/PhyxView.vue
+++ b/src/components/phyx/PhyxView.vue
@@ -83,7 +83,7 @@
                      @click="$store.commit('toggleCookieAllowed')"
               />
               Save curator information and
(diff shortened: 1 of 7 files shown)


hlapp (approved):
Looks good and is indeed welcome. Now that we're storing cookies in a client's web browser I also think we should at the very least say somewhere that we're doing this. But this can be a separate issue.

gaurav:
I've added a checkbox to the "Phyloreference collection" box that looks like this:

Screenshot 2022-11-14 at 10 58 45 PM

As described, the checkbox will either turn on cookies (by creating a COOKIE_ALLOWED cookie on the browser, which allows the getting and setting of cookies to proceed) or -- if unchecked -- delete all Klados cookies, including the COOKIE_ALLOWED cookie. Note that the curation information is in the old format; I'm working on an improved curation information format in https://github.com/phyloref/phyx.js/issues/119.

Here's the text if you would like to suggest improvements:
> Click this checkbox if you would like Klados to save your curator information and default nomenclatural code as a browser cookie for the next month, or uncheck it to delete all Klados cookies from your browsers.

src/components/phyx/PhyxView.vue
```diff
@@ -65,14 +65,28 @@
...
+
+          <!-- Checkbox for permission to save this information to browser -->
+          <div class="form-group row">
+            <div class="col-md-2">&nbsp;</div>
+            <div class="col-md-10">
+              <input type="checkbox"
+                     :checked="cookieCheckbox"
+                     @click="$store.commit('toggleCookieAllowed')"
+              />
+              Click this checkbox if you would like Klados to save your curator information and
```
hlapp, line 85:
```suggestion
              Save curator information and
```

src/components/phyx/PhyxView.vue
```diff
@@ -65,14 +65,28 @@
...
+              default nomenclatural code as a browser cookie for the next month, or uncheck it to
```
hlapp, line 86:
```suggestion
              default nomenclatural code as a browser cookie (currently until the next month). If changed to unchecked, 
```

src/components/phyx/PhyxView.vue
```diff
@@ -65,14 +65,28 @@
...
+              delete all Klados cookies from your browsers.
```
hlapp, line 87:
```suggestion
              Klados cookies will be deleted from your browser.
```

hlapp (approved):
Looks good, except I suggested some changes to the label for the checkbox. BTW your phrasing of _for the next month_ is ambiguous (open to various interpretations): for the next 30 days? until the end of the next month? until the end of this month? I chose one interpretation, but you should substitute  the correct and unambiguous one.

src/components/phyx/PhyxView.vue
gaurav: I've made this explicit as "(currently for thirty days)." in ddac2cd. However, we can easily change that to one month if that is preferable (as per https://www.npmjs.com/package/vue-cookies).

PR #266, opened 2022-10-04
Fixed text in "Actual resolved nodes" field in PhylorefView

### LINKED ISSUE #262: UI says "Click 'Reason'" but there is no UI element labeled 'Reason'
In the "Expected and actual resolution" box the following UI element says to click 'Reason'.

image

However, there is no such UI element labeled 'Reason', at least not in the deployed version (refs/tags/v0.3.1). Under Actions, there is one option named "Resolve against phylogenies", which if clicked appears to kick off reasoning. So perhaps this is the button being referred to? If so, this would hardly be apparent to anyone not intimately familiar with how "Resolve against phylogenies" would work internally.

gaurav:
This PR fixes the text in the "Actual resolved nodes" field in the PhylorefView, replacing the previous "Reason" to the new text of the button that triggers reasoning, "Resolve against phylogenies".

Screen Shot 2022-10-04 at 5 57 56 PM

Closes #262

### DIFF
--- a/src/components/phyloref/PhylorefView.vue
+++ b/src/components/phyloref/PhylorefView.vue
@@ -355,7 +355,7 @@
                     readonly
                     type="text"
                     class="form-control"
-                    value="Click 'Reason' to reason over all phyloreferences."
+                    value="Click 'Resolve against phylogenies' to reason over all phyloreferences."
                   >
                 </template>
                 <template v-else>


hlapp (approved):
👍 in the sense that this addresses the concrete issue. I also don't feel it's the most intuitive wording either (it's IMHO not intuitive to causally connect the ability of "Go to node" with first having to "Reason against phylogenies"), but we shall leave that to some time in the future.

PR #268, opened 2022-10-05
Replace cladeDefinition with definition

gaurav:
We previously used `cladeDefinition` as the field name of the verbatim clade definition in a phyloref, but we've since updated it to `definition`. This PR makes that change in Klados.


hlapp (approved):
👍

PR #274, opened 2022-11-02
Added a test for the reasoner backend

### LINKED ISSUE #273: Add scheduled test for the Klados backend
We recently ran into a problem where the Klados backend went down, which we only noticed because Hilmar tried to access it (#264). We should have some kind of is-system-up scheduled check to make sure that the backend is up and capable of responding to requests. Because of the way the backend is configured, we can't just check if the web server at https://reasoner.phyloref.org/reason is up -- we actually need to submit a small job for it to process, and confirm that it reasons correctly.

I've used Pingdom previously to do this kind of check, but Hilmar suggested GitHub Actions with a [schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) trigger, maybe running this check hourly and e-mailing us an alert when Klados goes down and when it comes back up.

gaurav:
Adds a GitHub Action scheduled to run twice a day (3am and 3pm UTC = 11am and 11pm EDT) that sends a submission to the Klados reasoner backend and passes the test only if it resolves it as expected.

Closes #273.


hlapp (approved):
Great, looks good! I assume you'll squash the commits into one for merging?

gaurav:
Sure, I can do that!

PR #278, opened 2022-11-30
Add additional example file: Fisher et al, 2007

### LINKED ISSUE #232: Add additional example files

gaurav:
This PR restores Fisher et al, 2007 from previous versions of Klados so that we can have examples of phyloreferences with specimens so we can make sure that Klados supports them. Closes #232.

It also includes some minor fixes that were discovered in creating this example file:
- We previously had separate "Primary reference phylogeny" and "Reference phylogeny" citations for phylogenies. This has now been replaced with a single "Source" citation as per the Phyx manuscript.
- Added the phyloref definition to the CSV export.
- Klados attempted to read the default nomenclatural code from the `defaultNomenclaturalCodeURI` field from the Phyx file, but the specification says this field should be called `defaultNomenclaturalCodeIRI`. Fixed.

The backend is down (I'm working on fixing it), but I have confirmed that this works locally on my computer:

Screenshot 2023-03-05 at 1 23 22 AM


gaurav:
Aha, I've figured out what the problem is with *both* of these files: the phylogeny nodes aren't being notated with any nomenclatural code. This is because phyx.js uses the following logic to determine how to choose a nomenclatural code:

https://github.com/phyloref/phyx.js/blob/d3623e3a97b3d46747295dd826b735aab7ed4f16/test/nomenclatural-codes.js#L116-L124

I should be able to use `defaultNomenclaturalCodeIRI` to set these, but (1) I'm curious as to why the second part of that algorithm didn't fix this as it does for `brochu_2003.json` -- I hope it's not because the occurrence entries confuse it!, and (2) I wonder why we didn't add a per-phylogeny field for storing the nomenclatural code -- that might be a neater solution than a per-document `defaultNomenclaturalCodeIRI` field?

gaurav:
Moving content from the PR description into an issue so we document some fixes:


Outstanding issues with [*Fisher et al*](https://doi.org/10.1639/0007-2745(2007)110[46:POTCWA]2.0.CO;2):
- [x] Phyloref Exodictyon doesn't resolve because it [appears to use](https://user-images.githubusercontent.com/23979/215397861-4cc1f40a-a57d-45a3-bce5-83ee872f2651.png) the type (species?) of genus _Exodictyon_ as an internal specifier, in addition to specimen `Wall 2527, Fiji (UC)`, which is identified elsewhere in the publication as _Exodictyon incrassatum_. What should we match this type to?
   - This means that the type species of the genus is included (even if not present in the phylogeny). So we should annotate _Exodictyon incrassatum_ as _Exodictyon_, and then both specifiers will resolve.
- [x] Phyloref Leucophanella doesn't resolve because it includes an internal specifier, `internal Specifier: Type: S. revolutus Dozy & Molk.`, which isn't included in the phylogeny (or anywhere else in the paper). I've deleted it and mentioned the deletion in the curator comments.

hlapp (approved):
👍

PR #285, opened 2023-02-01
Add taxonomic units to phylogeny view

gaurav:
This PR allows taxonomic units to be edited for a node.

It might be useful to also have some sort of automatic processing of genus names within a phylogeny, i.e. if a particular genus only appears once, we might assume that that node represents that genus on this phylogeny.

WIP


gaurav:
As far as I can tell, all the work on getting Fisher et al working again was merged in https://github.com/phyloref/klados/pull/278. I've moved the remaining commits into https://github.com/phyloref/klados/pull/307, so this PR has no unique code remaining in it and can be closed.

PR #288, opened 2023-03-08
Add citation

### LINKED ISSUE #284: Add suggested citation to README

gaurav:
This PR adds a CITATION.cff file for Klados as well as a suggested citation in README.md. It also adds a button to the Zenodo archive for this repository, cited with the DOI for the latest version (https://zenodo.org/record/7103131) instead of the Klados v0.4.0 release (https://zenodo.org/record/7103132).

Closes #284.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -39,6 +39,7 @@ repository-code: 'https://github.com/phyloref/klados'
 url: 'https://phyloref.org/klados/'
 repository: 'https://zenodo.org/record/7103132'
 abstract: >-
-  A curation tool to edit test cases for the
-  Phyloreferencing curation workflow 
+  Tool for curating phyloreferences side-by-side with the phylogenies they were originally defined on. Phyloreferences
+  and their accompanying phylogenies are stored in the Phyx format, which can be downloaded as an OWL 2 ontology in the
+  JSON-LD format.
 license: MIT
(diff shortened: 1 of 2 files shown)


CITATION.cff
```diff
@@ -0,0 +1,44 @@
...
+identifiers:
+  - type: doi
+    value: 10.5281/zenodo.7103131
+    description: Source code archived to Zenodo
+repository-code: 'https://github.com/phyloref/klados'
+url: 'https://phyloref.org/klados/'
+repository: 'https://zenodo.org/record/7103132'
+abstract: >-
+  A curation tool to edit test cases for the
+  Phyloreferencing curation workflow
```
hlapp, line 43: I don't know why we would use an abstract here that's much more narrow, and thus not fully accurate, compared the quite apt brief description you put into the README:

> Klados allows users to curate phyloreferences side-by-side with the phylogenies they were originally defined on. The curated phylogenies are stored in the Phyx format, which can be downloaded as an OWL 2 ontology in the JSON-LD format.

hlapp: If you wanted to make this non-redundant with the name (title), you could modify to something like the following:

> Tool for curating phyloreferences side-by-side with the phylogenies they were originally defined on. Phyloreferences and their accompanying phylogenies are stored in the Phyx format, which can be downloaded as an OWL 2 ontology in the JSON-LD format.

hlapp (changes requested):
See inline comment.

CITATION.cff
gaurav: Thanks, Hilmar! I've updated the abstract to your suggestion in 78f9c2c.

hlapp (approved):
👍🏻

PR #289, opened 2023-03-08
Add support for exporting annotated phylogenies

### LINKED ISSUE #226: Add support for exporting annotated phylogenies as Newick
Klados can display an annotated phylogeny (and show where a particular phyloreference resolves), but there is currently no way to export this phylogeny as a Newick string, allowing downstream use of the annotated phylogeny.

Note that Klados currently can only display a phylogeny annotated with a single phyloreference at a time. If we want to implement a view that includes all known annotations on a single phylogeny, this will need to be done in a separate issue.

gaurav:
This PR adds a button to phylogenies that allow them to be exported. Closes #226.

Screenshot 2023-03-08 at 1 36 26 AM

WIP:
- [x] The output includes a bunch of extra characters at the start of the file -- maybe the Unicode BOM? We need to get rid of those.
- [ ] Currently, this exports the entire Newick file, which is fine for the phylogeny view. For the phyloref view, however, we only want to display that specific phyloreference. Our current code filters out the other labels when drawing the phylogeny -- it might be worth looking at phylotree.js and seeing if there is some other way to hide internal node labels from both display as well as the Newick export.
- [ ] We also want to distinguish between expected phyloreferences (as labeled in the phylogeny) and resolved phyloreferences (as determined by reasoning). Our plan in #226 is to leave expected phylorefs as-is (e.g. `Alligatoridae`) and to add either `phyloref_` or `reasoned_` as a prefix when the node was found by the reasoner (e.g. `phyloref_Alligatoridae` or `reasoned_Alligatoridae`).

### DIFF
--- a/src/components/phylogeny/Phylotree.vue
+++ b/src/components/phylogeny/Phylotree.vue
@@ -22,6 +22,13 @@
         class="col-md-12 phylogeny"
       />
       <ResizeObserver @notify="redrawTree" />
+      <button
+        type="button"
+        class="btn btn-primary"
+        @click="exportAsNewick()"
+      >
+        Download as Newick
+      </button>
     </div>
   </div>
 </template>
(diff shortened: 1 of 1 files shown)


gaurav:
Deprecated in favor of PR #313, which uses the updated code. Closing this PR.

PR #291, opened 2023-03-08
Add Testudinata from phylonym as an example of an apomorphy-based phyloref

gaurav:
This PR adds support for entering apomorphy-based clade definitions in Klados. Note that the Testudinata citation isn't working because of https://github.com/phyloref/klados/issues/301 -- I think we can review this PR without fixing that. The remaining functionality appears to be working.

Initially the Phyloref interface looks like this:
Screenshot 2023-06-21 at 3 14 44 AM

Clicking on the box in the top right of the "Apomorphy" box causes the description to open:
Screenshot 2023-06-21 at 3 02 10 AM

This PR includes an apomorphy-based definition from Phylonym:
Screenshot 2023-06-21 at 3 16 25 AM

I've also modified the summary page to indicate that we don't expect apomorphy-based phylorefs to resolve:
Screenshot 2023-06-21 at 3 01 15 AM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "filesaver.js-npm": "^1.0.1",
         "jquery": "^3.5.1",
         "lodash": "^4.17.19",
-        "newick-js": "^1.1.1",
+        "newick-js": "1.1.1",
         "popper.js": "^1.16.1",
         "vue": "^2.6.11",
         "vue-cookies": "^1.8.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -5205,9 +5205,9 @@
       }
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001409",
-      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001409.tgz",
-      "integrity": "sha512-V0mnJ5dwarmhYv8/MzhJ//aW68UpvnQBXv8lJ2QUsvn2pHcmAuNtu8hQEDz37XnA1iE+lRR9CIfGWWpgJ5QedQ==",
+      "version": "1.0.30001486",
+      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001486.tgz",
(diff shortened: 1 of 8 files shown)


gaurav:
- [ ] Add a checkbox after the word "Apomorphy" in the title
- [ ] Leaning towards removing the "Bearing entity" and "Phenotypic quality" fields -- for this version, just the free text definition is probably sufficient.
- [ ] Explicitly include PATO and UBERON terms in the free text example so that users know that they can include term URIs or Manchester syntax in the freetext definition if necessary.

In the future, we might want a "related entity" field as well. Gaurav to talk to Jim about this. Maybe a Manchester syntax field?

gaurav:
Here's where I am with the interface right now (as of a1d9cdd). I moved all the specifier groups into cards, which makes it easier to add elements to the header. Adding a check box immediately to the right of the "Apomorphy" looked odd, but then I could realize I could add a button to float right in all three card headers -- for "Internal specifiers" and "External specifiers", this button would add a new specifier to the list, and for "Apomorphy" it would toggle whether any apomorphy was present. So, if there is no apomorphy, the specifier card looks like this:

Screenshot 2023-04-26 at 1 47 01 AM

But if you then click on the check box beside the "Apomorphy", the apomorphy check box is selected and the apomorphy details appear:

Screenshot 2023-04-26 at 1 46 33 AM

I'm looking through the [Bootstrap icons](https://bootstrap-vue.org/docs/icons#icons-1) to see if I can find one that more clearly indicates "This is a checkbox". If I can't find one, I'm planning to change the text in the apomorphy to read:

> No apomorphy in this phyloreference. [Set one?](https://example.org)

Where clicking on "Set one?" would select the checkbox and display the apomorphy details.

What do y'all think?

hlapp:
I think that's reasonable. As we discussed, I wouldn't spent too much time on over optimizing the UI.

hlapp (approved):
Looks overall good. I notice that there are repeated instanced of `id=bearing-entity` in the VUE code, but perhaps I'm just not understanding VUE well enough. So I'll tentatively approve.

gaurav:
@hlapp Good catch -- I did have some misnamed ids! They don't matter for rendering, but I've fixed them in 1ac9ec1.

PR #295, opened 2023-05-24
Fix CITATION.cff

### LINKED ISSUE #293: CITATION generated citation different (and less informative) than the one given in README
It seems the reason for this is that the `title` field is simply "Klados":
https://github.com/phyloref/klados/blob/a987851221320d8b357d334febfa05d53bb2c0a9/CITATION.cff#L5

Whereas we want this to be *Klados: a tool for authoring, testing and curating phyloreferences*.

The other thing I notice is that the generated BibTeX doesn't include the Zenodo DOI, which seems both unfortunate and undesirable. Any idea how that might be fixed?

gaurav:
This PR attempts to fix CITATION.cff as reported in #293. I can't figure out how to test this in GitHub without merging it, but I tried running `cffconvert` on it locally and it fixed both of the reported issues.

```
❯ cffconvert -i CITATION.cff -f apalike
Vaidya G., Becker A., Cellinese N., Lapp H. Klados: a tool for authoring, testing and curating phyloreferences DOI: 10.5281/zenodo.7103131 URL: https://phyloref.org/klados/
```

Once we merge this, I'll double-check that it fixed #293 and then close that issue.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -2,7 +2,7 @@
 # Visit https://bit.ly/cffinit to generate yours today!
 
 cff-version: 1.2.0
-title: Klados
+title: "Klados: a tool for authoring, testing and curating phyloreferences"
 message: >-
   If you use this software, please cite it using the
   metadata from this file.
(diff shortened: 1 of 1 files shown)


hlapp (approved):
👍

Issue #71, opened 2018-06-27
Remind users to note whether a clade definition specifies a crown clade in the verbatim definition


gaurav:
In an embargoed curated example, we've found a clade definition whose verbatim definition starts with "The largest crown clade containing ...", even though the clade definition points to the [total or pan group](https://en.wikipedia.org/wiki/Crown_group#Pan-group). Readers are therefore expected to resolve the clade definition, and then exclude the stem groups that would otherwise be included, in order to obtain the [crown group](https://en.wikipedia.org/wiki/Crown_group#Pan-group).

The Curation Tool doesn't currently have a way of indicating that a phyloreference should be restricted to the crown clade, i.e. the reasoner should try to exclude stem groups automatically after resolving the clade definition. This property will require additional annotation of the phylogeny to indicate which terminal groups are extinct, so this issue covers adding that annotation as well.

ncellinese:
I think you are confused here, or I am not sure I understand what you mean. A pan group or total clade is essentially a branch based clade that includes a crown clade rather than just any clade (i.e. a clade with fossils). So what’s the issue here?

Nico

gaurav:
Nico and I just had an in-person discussion about this, and decided that this works better with an example. The actual example comes from clade Angiospermae. In the reference phylogeny, pan-Angiospermae includes four extinct lineages (Glossopterids, Pentoxylon, Bennettitales, Caytonia) and one extant lineage (Angiospermae). If the definition were simply "The largest clade containing ... but not ...", it would resolve to pan-Angiospermae. But since it is "The largest _crown_ clade containing ... but not ...", we need to exclude the four extinct lineages so that we instead resolve to Angiospermae. If we have some way of annotating which terminal nodes are extinct in the phylogeny, we could exclude them after reasoning.

In addition, the clade definition of pan-Angiospermae is simply "the total clade of the crown clade Angiospermae". So, whatever mechanism we develop for excluding extinct lineages, we have to be able to subsequently ignore them if necessary. This would be easiest if carried out after reasoning, but we could also try to find ways to integrate this into the OWL restrictions we use.

gaurav:
The information that a phyloreference should be resolved as a crown clade should probably be included in the curator notes (as we do for the reference taxon information, #70), since this information is not used to construct the logical expressions in OWL in any way. Therefore, I propose that we explicitly note that crown clade status should be included in that field and then close this issue.

hlapp:
We should discuss but I think I agree.

[...]

Issue #218, opened 2021-07-08
Improve display of the placeholder text for "Curator Comments"


gaurav:
The "Curator Comments" field has a placeholder text to provide an example of the type of curator comments that can be entered there. However, it is dark enough that it can be confused for actual text in that box. This should be rectified, or moved outside the box in some way.

gaurav:
This would require modifying the CSS to change the placeholder settings on a per-browser basis (https://stackoverflow.com/a/32815476/27310).

gaurav:
- [ ] Just starting with "e.g." should clarify that this is a placeholder text.
- [ ] Phrase it like an action, "Enter the ..."
- [ ] Get rid of unnecessary placeholders, e.g. "Phylogenetic clade definition"

gaurav:
I think the curator notes placeholder text is now clear enough, even without adding "Enter the ...".

<img width="888" alt="Screenshot 2023-05-24 at 3 31 14 AM" src="https://github.com/phyloref/klados/assets/23979/b44e4f1f-1a94-4358-943a-5c9774e96ba5">

gaurav:
I think this issue should be fully closed by the previous work on improving the curator notes placeholder text and by removing redundant placeholders in https://github.com/phyloref/klados/pull/297.

PR #297, opened 2023-05-24
Remove unnecessary placeholder text

gaurav:
When initially building the Klados user interface I added a lot of placeholder texts. This PR removes placeholders texts that are redundant (as per https://github.com/phyloref/klados/issues/218) or moves the non-redundant text into the field label. It does improve the placeholder for the clade definition to `e.g. 'The largest crown clade containing Canis lupus Linnaeus 1758 (Canidae) but not Felis silvestris Schreber 1777 (Felidae).'`, which should also close https://github.com/phyloref/klados/issues/71.

Screenshot 2023-05-24 at 3 58 55 AM


hlapp (approved):
Looks all good!

hlapp (approved):
Looks all good!

PR #299, opened 2023-06-18
Upgraded node packages using `npm upgrade`

gaurav:
This PR upgrades Node packages using `npm upgrade`. Because of https://github.com/keesey/newick-js/issues/4, we need to pin `newick-js` to v1.1.1 for now.


hlapp (approved):

hlapp:
Wait - doesn't keesey/newick-js#4 say
> Install 1.2.1 -- that should do it.

Rather than 1.1.1?

gaurav:
I tried to upgrade newick-js to v1.2.1 in Node 16, but ran into some errors:

```
Module parse failed: Unexpected token (4:14)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| Object.defineProperty(exports, "__esModule", { value: true });
| class GraphWriter {
>     rootWeight;
|     constructor(graph, rootWeight = NaN) {
|         this.rootWeight = rootWeight;
```

I will need to investigate this further, although note that this error doesn't appear in PR https://github.com/phyloref/klados/pull/303.

gaurav:
I figured this out this issue in PR #303, so this PR is no longer needed. Closing.

PR #304, opened 2023-08-02
Delete duplicated citation code

### LINKED ISSUE #301: Replace citation-to-text code with that provided by phyx.js
We developed the citation-to-text handling code in Klados before moving it to phyx.js. We should fix this duplication by deleting the duplicate code in Klados and calling the appropriate methods in phyx.js instead.

https://github.com/phyloref/klados/blob/ce4e31b1de251ff10a22cbc5b2e17ceb41f73436/src/store/modules/citations.js#L115-L118

gaurav:
Two pieces of code were copied over from Klados to Phyx.js at some point:
1. The code for [generating a single string](https://github.com/phyloref/phyx.js/blob/86878b987b07254068a5e7ca372e32d0d0174881/src/wrappers/CitationWrapper.js#L38) to describe a citation.
2. The code for [generating a single string](https://github.com/phyloref/phyx.js/blob/86878b987b07254068a5e7ca372e32d0d0174881/src/wrappers/CitationWrapper.js#L22-L35) for an "agent name" (i.e. for an author, editor or series editor).

This PR removes the Klados version of this code, and calls the phyx.js version of this code instead. This avoid duplication, which is a particular problem as the two pieces of code had slightly different ways of parsing agent names as well as some other fields. Closes #301.

Should be merged after PR #303.

### DIFF
--- a/src/components/citations/Citation.vue
+++ b/src/components/citations/Citation.vue
@@ -626,7 +626,7 @@ export default {
     },
     wrappedCitation(citation) {
       // Return the citation
-      return this.$store.getters.getWrappedCitation(citation);
+      return this.$store.getters.getCitationModel(citation);
     },
   },
 };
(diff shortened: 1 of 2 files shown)


hlapp (approved):
👍

PR #305, opened 2023-08-02
Add phylotree.js as an NPM package

### LINKED ISSUE #258: Update to latest version of Phylotree.js

gaurav:
This PR replaces an old version of phylotree.js that we included with Klados with [phylotree v1.4.0](https://www.npmjs.com/package/phylotree) installed from NPM.js.

Closes #258.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,12 +6,6 @@
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="icon" href="/favicon.ico">
 
-    <!-- Include Phylotree.js -->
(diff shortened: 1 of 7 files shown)


gaurav:
Current status:

Screenshot 2023-08-16 at 4 32 10 AM

hlapp (approved):
As we discussed, the distance/branch length scale being displayed even though there are no branch lengths is undesirable, but would almost certainly require "hacks" to bring under our control, and arguably it isn't causing any real harm. Hence, 👍

PR #306, opened 2023-09-27
Fix borders on sidebar

gaurav:
This PR incorporates the sidebar fixes from https://github.com/phyloref/klados/pull/185 and extends them to fully fix the sidebar borders. It also removes the "Add specifier" button, which is no longer functional or necessary.

Screenshot 2023-09-26 at 11 50 25 PM

Supercedes PR https://github.com/phyloref/klados/pull/185


hlapp (approved):
👍

PR #307, opened 2023-09-27
Modify taxonomic units on phylogenies

### LINKED ISSUE #228: Implement support for editing multiple taxonomic units at a single phylogeny node
There are three use-cases in which a single phylogeny node may need to be recognized by the software as referring to multiple taxonomic units, with examples from the Fisher et al, 2007 example included in Klados:
1. When a node is a synonym of a specifier
   - In Fisher et al, 2007: phyloref Mitthyridium has *Codonoblepharum undulatum* as an internal specifier, but the phylogeny uses the name *Mitthyridium undulatum*. For reasoning to work, this node needs to be identified as *M. undulatum*, but from a provenance perspective, it would be incorrect to simply rename this node. The best solution is for this node to have two taxonomic units: *C. undulatum* and *M. undulatum*.
2. When a node includes a specifier
    - In Fisher et al, 2007: phyloref Exodictyon has the genus *Exodictyon* excl. parte, emend Ellis in Lindbergia 11: 16 (1985) as an internal specifier. However, the phylogeny only refers to *Exodictyon incrassatum*. We therefore need to attach another taxonomic unit to *E. incrassatum* on the phylogeny to indicate that the genus should also be matched at this node. Note that the reverse case is more common: a phyloreference usually cites species or specimens as its specifiers, but sometimes the phylogeny only goes down to the genus or family level.
3. To suppose specimen identifiers
   - In Fisher et al, 2007: phyloref Leucophanes has an external specifier defined as specimen Wall 2527, Fiji (uc). By looking up information about this specimen in the paper, we can determine that it was identified as *Exodictyon incrassatum*, which is only present once in the phylogeny. This must be Wall 2527, Fiji (uc). However, Newick does not support both specimen identifiers and taxon names at a single node (except by creating our own format or by using Newick comments). Instead, we should create a taxonomic unit for this specimen and add it to the *E. incrassatum* node on the phylogeny to support this feature.

While the underlying data model allows us to associate multiple taxonomic units with a single phylogeny node, the UI currently only allows the user to rename individual phylogeny nodes. This means that a user can't annotate a single node as both a species and a genus (if it is the only representative of a genus in the phylogeny, say -- see #45), and would require a more complex system for recording relationships between taxa (#129).

Instead, I propose that we modify the UI so that:
1. A user can right-click on a node and choose to either "Rename this node" or "Modify the taxonomic unit at this node" (alternatively: "Modify the 4 taxonomic units at this node", etc.).
5. This will display a dialog box that will display all taxonomic units associated with this node. The user can add, edit or delete any taxonomic unit. We might add support for a citation or curator notes if the user would like to note why they believe a particular taxonomic unit matches a particular node.
6. To both the Phyloreference and Phylogeny view, add a read-only table that displays a list of all phylogeny nodes that explicitly match taxonomic units (i.e. nodes that don't implicitly have a taxonomic unit extracted from their label).

I think this will satisfy our most common use-cases:
1. A user can indicate that on a particular node on a particular phylogeny matches several taxonomic units, i.e. a single node might represent *Homo sapiens*, clade *Homo* and clade Hominoidea. This would allow a phylogeny whose leaf nodes represent clade *Hominoidea* to resolve a phylogeny that refers to *Homo sapiens* as a specifier.
2. These relationships will be displayed on the Phylogeny page as well as the Phyloreference page, so it will be clear to users what these relationships are. This will allow us to meet the needs of #129 without the complex data model described in #130.

One good/bad thing about this feature is that these taxon unit relationships will be specific a particular phylogeny. #130 proposes an alternate approach, in which we record taxonomic relationships separately in the Phyx file and then apply it to all specifier matching within the file. In this case, we could record that genus *Homo* always includes *Homo sapiens* (citing a taxonomic source for this), and then apply this to any phylogeny that was included in that Phyx file. However, I think that level of complexity is unnecessary for our needs, and so I propose we close #130 and replace it with this implementation (or postpone both implementations until after Klados v1.0).

gaurav:
This PR adds support for modifying the taxonomic units represented by a node in a phylogeny. This functionality has existed in Phyx.js for a while now, so it's really just a case of surfacing it in the UI with minimum work but in a way that's easy to work with.

Closes #228

This combines the work started on PR https://github.com/phyloref/klados/pull/203 and PR https://github.com/phyloref/klados/pull/285.

Current status:
- [x] Existing taxonomic edits can be edited 
- [x] Add support for adding new taxonomic units
- [x] Add UI support for deleting taxonomic units

Screenshot 2023-10-25 at 1 41 15 AM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3021,9 +3021,9 @@
       }
     },
     "node_modules/postcss": {
-      "version": "8.4.24",
-      "resolved": "https://registry.npmjs.org/postcss/-/postcss-8.4.24.tgz",
-      "integrity": "sha512-M0RzbcI0sO/XJNucsGjvWU9ERWxb/ytp1w6dKtxTKgixdtQDq4rmx/g8W1hnaheq9jgwL/oyEdH5Bc4WwJKMqg==",
+      "version": "8.4.31",
+      "resolved": "https://registry.npmjs.org/postcss/-/postcss-8.4.31.tgz",
+      "integrity": "sha512-PS08Iboia9mts/2ygV3eLpY5ghnUcfLV/EXTOW1E2qYxJKGGBUtNjN76FYHnMs36RmARn41bC0AZmn+rR0OVpQ==",
       "funding": [
         {
           "type": "opencollective",
--- a/src/components/phylogeny/PhylogenyView.vue
+++ b/src/components/phylogeny/PhylogenyView.vue
@@ -151,15 +151,22 @@
         Taxonomic units in this phylogeny
       </h5>
       <div class="card-body">
-        <b-table striped hover filter :items="taxonomicUnitsTable" :primary-key="node_label" show-empty>
-          <template #empty="scope">
-            <h4>No labels found in this phylogeny.</h4>
-          </template>
-          <template #emptyfiltered="scope">
-            <h4>No labels found after filtering.</h4>
-          </template>
-        </b-table>
+        To add additional taxonomic units to this list, please label the corresponding node on the phylogeny.
       </div>
+      <b-table
+        striped
+        hover
+        :items="taxonomicUnitsTable"
+        :primary-key="node_label"
+        show-empty
+      >
+        <template #empty="scope">
+          <h4>No labels found in this phylogeny.</h4>
+        </template>
+        <template #emptyfiltered="scope">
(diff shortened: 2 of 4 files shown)


hlapp (approved):
LGTM as far as I can tell.

PR #308, opened 2023-09-27
Support phylogenies and phyloreferences that cannot be parsed

### LINKED ISSUE #187: Make sure an incorrect phylogeny doesn't break the rest of the UI
Anna appears to have found a case where a malformed Newick string causes the phyloreference summary page to blank out because of an internal error. We should trap this and make sure the page can be rendered even with a malformed Newick string.

### LINKED ISSUE #196: Additional checks for malformed Newick strings
Anna ran into some problems with entering some Newick strings into Klados, especially since we report the following error from our Newick parser when they show up: "**Error parsing phylogeny.** An error occured while parsing this phylogeny: End of buffer reached."

It would help if we checked for some common errors that she has identified:
- [x] Sometimes two or more phylogenies get concatenated together (e.g. `(A, (B, C) (D, (E, F))` or `(A, (B, C)); TREE tree3 = (D, (E, F))`, which cannot be read as a single phylogeny. This is hard to detect as phylogenies can be long and complicated. We could look for text after an initial phylogeny to see if there's anything missing.
- [x] If unbalanced quotes are used in the phylogeny, this can cause an "end of buffer reached" error as the parser is looking for the terminating quote. Weirdly enough, this can even show up when the single quote is surrounded by double-quotes (i.e. `"Steller's_sea_lion"` will cause an error). We could detect and report this separately.

Additional Newick parsing issues can be added here or separated into different issues.

gaurav:
Phylogenies and phyloreferences can throw exceptions when you attempt to load them or convert them into JSON-LD/n-Quads, which we've previously ignored. This PR adds `try {} catch {}` blocks around such invocations so that we can produce an error instead of having the entire Vue app crash. It also adds the download filename to the Advanced Modal dialog download button for completeness.

Also includes a minor fix to the formatting of a few lines.

Closes #187.

Also checks for unbalanced single or double quotes in the Newick file as illustrated in this screenshot:

Screenshot 2024-02-27 at 11 03 30 PM

Closes #196.

### DIFF
--- a/src/components/modals/AdvancedOptionsModal.vue
+++ b/src/components/modals/AdvancedOptionsModal.vue
@@ -115,11 +115,15 @@ export default {
       // Exports the PHYX file as an OWL/JSON-LD file, which can be opened in
       // Protege or converted into OWL/XML or other formats.
       const wrapped = new PhyxWrapper(this.$store.state.phyx.currentPhyx);
-      const content = [JSON.stringify([wrapped.asJSONLD()], undefined, 4)];
+      try {
+        const content = [JSON.stringify([wrapped.asJSONLD()], undefined, 4)];
 
-      // Save to local hard drive.
-      const jsonldFile = new File(content, 'download.jsonld', { type: 'application/json;charset=utf-8' });
-      saveAs(jsonldFile);
+        // Save to local hard drive.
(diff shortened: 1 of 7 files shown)


hlapp (approved):
LGTM! 👍

PR #310, opened 2023-10-11
Update to v0.5.1

gaurav:
I forgot to increment the version number to [v0.5.0](https://github.com/phyloref/klados/releases/tag/v0.5.0) when I created that tag and release. This PR fixes that (but goes to v0.5.1 instead) and also upgrades packages using `npm upgrade`.

It looks like the Vite upgrade broke [our GitHub publish action](https://github.com/phyloref/klados/actions/runs/6478461467/job/17590255855) because it writes the website to `dist/` instead of `docs/`. I've fixed this in b3e00e0.

### DIFF
--- a/.github/workflows/deploy-to-github-pages.yml
+++ b/.github/workflows/deploy-to-github-pages.yml
@@ -12,7 +12,6 @@ name: Deploy to GitHub Pages
 #     key. This key is stored as an encrypted secret in the repository.
 
 on:
-  push:
   release:
     types: [published]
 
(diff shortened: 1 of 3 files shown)


hlapp:
Just to confirm, the putting back into draft mode wasn't accidental, but presumably because you found the broken GitHub action issue. So I'll stand back until re-assigned 😊

gaurav:
Correct. Although I might have already figured out what's wrong, so it might be back with you pretty soon!

ncellinese:
Alright Guys! Thank you! I don’t think we need to meet today. Better work on things instead :-)

Cheers, Nico

hlapp (approved):

PR #312, opened 2023-11-20
Upgrade phylotree.js and fix node renaming

### LINKED ISSUE #197: Bug: when renaming a node, spaces are ignored
Inserting an annotation indicating the expected node for a phyloreference is tricky to do directly in Newick. Clicking on a node in Klados opens a menu, which Phylotree.js allows us to add custom menu items to. We previously had a menu item for renaming a node, which would allow us to annotate nodes much more easily. We should reimplement this menu item to make it easier for curators to indicate which node a phyloreference should resolve to.

### LINKED ISSUE #200: Bug: when exporting a phylogeny as Newick, spaces in leaf node labels are ignored
There's an odd bug that comes up when using Phylotree to export its phylogeny as Newick: while internal node labels export fine, leaf node label become concatenated with themselves (e.g. "Mitthyridium undulatum" would become "Mitthyridium undulatumMitthyridiumundulatum"). We currently work around this ourselves, but we should come up with a minimum example that shows this issue and post an issue to Phylotree so they can fix it.

gaurav:
This PR upgrades all the NPM packages, which results in phylotree.js being upgraded to [phylotree.js v1.4.1](https://github.com/veg/phylotree.js/releases/tag/v1.4.1), which includes two changes we requested. This PR will also fix the issues related to that change in Klados. We also remove the escaping code I added previously since phylotree.js fixed this in PR https://github.com/veg/phylotree.js/pull/448.

Closes #197. Closes #200.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
         "lodash": "^4.17.19",
         "newick-js": "^1.2.1",
         "pako": "^2.1.0",
-        "phylotree": "^1.4.0",
+        "phylotree": "^2.0.1",
         "popper.js": "^1.16.1",
         "vue": "^2.7.7",
         "vue-cookies": "^1.8.1",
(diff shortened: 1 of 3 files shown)


hlapp (approved):
👍

PR #313, opened 2023-12-06
Add an option to export phylogenies in Nexus

### LINKED ISSUE #226: Add support for exporting annotated phylogenies as Newick

### LINKED ISSUE #290: Display multiple resolved phylorefs on a single phylogeny
We currently have two ways of displaying a phylogeny:
1. In the phylogeny view, we display the phylogeny as entered: we display ALL internal nodes.
2. In the phyloref view for a particular phyloref, we ignore all internal nodes except the node where the phyloreference was supposed to resolve (which is labeled with the node label from the phylogeny). We also highlight the internal node where the phyloref resolved, which might not be the same node -- however, we don't display any label there.

We should add one additional view to meet one of the use-cases described in the Klados manuscript ("Klados can display all the phyloreferences that resolve against a single phylogeny and export that as an annotated Newick file, which can be used for downstream phylogenetic analyses."): a single phylogeny in the Phyx view that displays (1) the entire phylogeny without any internal labels, but (2) with any resolved phyloreferences (maybe referring to them as e.g. `resolved_Alligatoridae` or `phyloref_Alligatoridae`).

If multiple phylorefs resolve to the same node, we can concatenate them with `_and_` (e.g. `phyloref_Alligatoridae_and_Alligatoroidea`).

If a phyloref resolves to a leaf node, we can add the phyloref name to the leaf node name (e.g. `resolved_Caimainidae_and_Caiman_crocodilus`).

gaurav:
This PR adds a "Download as Nexus" button below every phylogeny, both in the phyloreference view as well as the phylogeny view. This allows you to download that phylogeny as a Newick file.

Screenshot 2023-12-06 at 12 00 09 AM

On pressing the download button, you end up with a Nexus file that looks like this:

```
#NEXUS

BEGIN TREES;
  TREE klados_tree = ('Parasuchia':1,('rauisuchians':1,'Aetosauria':1,('sphenosuchians':1,('protosuchians':1,('mesosuchians':1,('Hylaeochampsa':1,'Aegyptosuchus':1,'Stomatosuchus':1,('Allodaposuchus':1,('Gavialis gangeticus':1,(('Diplocynodon ratelii':1,('Alligator mississippiensis':1,'Caiman crocodilus':1)'Alligatoridae':1)'Alligatoroidea':1,('Tomistoma schlegelii':1,('Osteolaemus tetraspis':1,'Crocodylus niloticus':1)'Crocodylinae':1)[&phyloref:actual="#Crocodylidae",phyloref:actualLabel="Crocodylidae"]:1)'Brevirostres':1):1)'Crocodylidae'[&phyloref:expected="#Crocodylidae",phyloref:expectedLabel="Crocodylidae"]:1)'Eusuchia':1)'Mesoeucrocodylia':1)'Crocodyliformes':1)'Crocodylomorpha':1):1):0;
END;
```

This format is supported by FigTree:

Screenshot 2024-01-16 at 1 11 16 AM

Should be merged after PR #312.

Closes #226. Closes #290.

Initial code copied from PR https://github.com/phyloref/klados/pull/289.

### DIFF
--- a/src/components/phylogeny/Phylotree.vue
+++ b/src/components/phylogeny/Phylotree.vue
@@ -16,8 +16,15 @@
     <div v-else class="phylotreeContainer">
       <div :id="'phylogeny' + phylogenyIndex" class="col-md-12 phylogeny" />
       <ResizeObserver @notify="redrawTree" />
-      <button type="button" class="btn btn-primary" @click="exportAsNewick()">
-        Download as Newick
+      <button
+        type="button"
+        class="btn btn-primary"
+        @click="exportAsNexus()"
+        data-toggle="tooltip"
+        data-placement="bottom"
+        title="Download Nexus file with annotations indicating phyloreferences."
+      >
+        Download as Nexus
       </button>
     </div>
   </div>
@@ -134,8 +141,8 @@ export default {
     this.redrawTree();
   },
   methods: {
-    exportAsNewick() {
-      // Export this phylogeny as a Newick string in a .txt file for download.
+    exportAsNexus() {
+      // Export this phylogeny as a Nexus string in a .nex file for download.
       const newickStr = this.tree.getNewick((node) => {
         // Is the resolved node for this phyloref? If so, let's make an annotation.
         if (
@@ -146,10 +153,9 @@ export default {
           const annotations = [];
           const data = node.data;
 
-          const removeColonsAndEquals = (str) => {
-            // Both colons and equal-tos have special significance in NHX
-            // annotations, so we need to escape them.
-            return str.replaceAll(":", "_").replaceAll("=", "_");
+          const convertToNexusAnnotationValue = (str) => {
+            // We really just need to wrap this in double-quotes, which means we need to filter out existing double quotes.
+            return '"' + str.replaceAll('"', "''") + '"';
           };
 
           if (
@@ -159,21 +165,21 @@ export default {
           ) {
             if (has(this.phyloref, "@id")) {
               annotations.push(
-                "klados_actual_id=" +
-                  removeColonsAndEquals(this.phyloref["@id"])
+                "phyloref:actual=" +
+                  convertToNexusAnnotationValue(this.phyloref["@id"])
               );
             }
 
             if (has(this.phyloref, "label")) {
               annotations.push(
-                "klados_actual_label=" +
-                  removeColonsAndEquals(this.phyloref["label"])
+                "phyloref:actualLabel=" +
+                  convertToNexusAnnotationValue(this.phyloref["label"])
               );
             }
 
             // We don't know what to call this phyloref, but nevertheless we label it minimally.
             if (!has(this.phyloref, "@id") && !has(this.phyloref, "label"))
-              annotations.push("klados_actual_id=");
+              annotations.push("phyloref:actual=");
           }
 
           if (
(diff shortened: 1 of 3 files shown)


gaurav:
- [ ] The key values need to be standardized in some way (we can come up with something to use instead of colons) and then they should be either reused from the Phyloref ontology or added to the Phyloref ontology.
- [ ] Order of branch length and annotation
  - Archaeopteryx (https://github.com/cmzmasek/archaeopteryx-js) by C. Zmasek, the same person who came up with the NHX.
  - Need to make sure that the Newick we produce can be read by simple Newick clients (i.e. those that don't understand the annotations)
  - For now, don't plan to 
- We currently export all internal labels. Do we really want to do that? NO
  - [ ] We should make it clear that the "Download as Newick with annotations" or "Download as Newick (plain)" -- we probably need some kind of documentation to explain what the annotations are. Question mark with pop-up: Make it clear what they will be getting: all the internal labels, NHX annotations indicating where the resolution.
- "shows every phyloreference and its resolution simultaneously": YES if it's not too complicated (which it might not be).

gaurav:
Some observations from trying this with FigTree v1.4.4:
- It doesn't appear to support annotations in Newick files, but does support them in Nexus files. So we might want to change our output format to Nexus.
- It doesn't need the `NHK` prefixes at all, and instead uses the format `'Alligatoridae'[&klados_actual_id="#Alligatoridae",klados_actual_label="Alligatoridae",klados_expected_id="#Alligatoridae",klados_expected_label="Alligatoridae"]:1`
- Once you use that format, you can do interesting stuff with it, such find it, color it, select it, and so on.
- Keys can be CURIEs (e.g. `phyloref:expected`) or URLs (e.g. `http://ontology.phyloref.org/phyloref.owl#expected`) and values can be URLs (e.g. `"http://ontology.phyloref.org/demo/Tetrapoda.owl#Amniota"`).

I also started playing around with Archeopteryx, but have been unable to display annotations as yet.

gaurav:
I've modified this PR to export Nexus files that are supported by FigTree! I think it's ready for review.

gaurav:
I've started a question on what we should call the `phyloref:expected[Label]`/`phyloref:actual[Label]` keys at https://github.com/phyloref/phyloref-ontology/issues/55

hlapp:
@gaurav I'm not sure but it seems like the plan is for this PR is to keep it open until phyloref/phyloref-ontology#55 is "resolved" and closed? Or is useful to treat them independent of each other?

gaurav:
@hlapp I think it makes sense to treat this PR separately -- here we can evaluate if this sort of annotation (and the associated UI) makes sense, and in the other issue we can figure out what final form that annotation should take. But I'm happy with other options (e.g. marking this PR as approved but making a note that we don't want to merge it until the final annotation has been decided).

src/components/phylogeny/Phylotree.vue
```diff
@@ -16,6 +16,16 @@
     <div v-else class="phylotreeContainer">
       <div :id="'phylogeny' + phylogenyIndex" class="col-md-12 phylogeny" />
       <ResizeObserver @notify="redrawTree" />
+      <button
+        type="button"
+        class="btn btn-primary"
+        @click="exportAsNexus()"
+        data-toggle="tooltip"
+        data-placement="bottom"
+        title="Download Nexus file with annotations indicating phyloreferences."
```
hlapp, line 25:
```suggestion
        title="Download Nexus file with annotations of where phyloreferences resolve"
```

hlapp: Just to make this a little shorter since it's a button.

src/components/phylogeny/Phylotree.vue
```diff
@@ -130,6 +141,91 @@ export default {
...
+          this.phyloref !== undefined &&
+          has(node, "data") &&
+          has(node.data, "@id")
+        ) {
+          const annotations = [];
+          const data = node.data;
+
+          const convertToNexusAnnotationValue = (str) => {
+            // We really just need to wrap this in double-quotes, which means we need to filter out existing double quotes.
+            return '"' + str.replaceAll('"', "''") + '"';
```
hlapp, line 158: So I take it there's no escape character? It bothers me a bit that we're substituting one character (double quote) with two other characters (two single quotes). This would look right if we output TeX, but we don't (nor is anyone necessarily expecting TeX), and it'll make for an essentially non-reversible substitution. (E.g., if a single quote were to be adjacent to a double quote, we'd have 3 single quotes and now on trying to reverse the substitution we wouldn't be able to automatically decide whether the first or the second two single quotes needed to be substituted back to a double quote.)

hlapp (approved):
See inline comments.

src/components/phylogeny/Phylotree.vue
gaurav: True! I could replace it with a single quote if that would be better. I tried the standard quote characters (`\"`, `""`) and neither of them worked. Do you have any ideas of other quote characters I could try?

gaurav: TODO: replace with a single quote.

gaurav: TODO: Provide some sort of warning message if there are already single quotes, "This cannot be substituted back -- do you want to proceed anyway?" -- that way users will know what to expect.

PR #314, opened 2024-01-03
Prevent phylorefs from being modified by being opened by the Specifier component

### LINKED ISSUE #263: Simply opening a phyloreference marks it as modified
At least in the deployed version (refs/tags/v0.3.1), when I open Brochu 2003 as the example file, then click on phyloreference Crocodylia, it immediately marks as modified, displaying the exclamation point marker and the "This phyloreference has been modified since being loaded! Use 'Save' to save your changes." banner.

gaurav:
WIP: I could try to create a complicated normalization method here in Klados, but the right way to do this is probably to do this piece-by-piece in phyx.js -- so that the PhyxWrapper tells the PhylogenyWrapper and PhylorefWrapper to wrap their respective objects, and so on down the line. This is being tracked there as https://github.com/phyloref/phyx.js/issues/134.

Should fix #263.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
         "@fortawesome/fontawesome-svg-core": "^1.2.29",
         "@fortawesome/free-solid-svg-icons": "^5.13.1",
         "@fortawesome/vue-fontawesome": "^0.1.10",
-        "@phyloref/phyx": "1.2.0-alpha1",
+        "@phyloref/phyx": "1.2.0-alpha2",
         "bootstrap": "^4.5.0",
         "bootstrap-vue": "^2.15.0",
         "buffer": "^6.0.3",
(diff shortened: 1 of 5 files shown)


gaurav:
- [ ] Make collapsed form non-editable and replace the "Expand" button with an "Expand/Edit" button or two buttons that do the same thing. Note that the edit button could be an icon rather than a full button.
- Change the logic: don't update the verbatim specifier, but do update the binomial/genus/specific epithet/infraspecific epithet if any of them change. That should get rid of this issue.
- If the file is missing details (e.g. no infraspecific epithet), which the tool fills in when you load the file.
    - We could highlight the specific modified field to indicate to the user that something here has changed.
    - Note that we can do this in the

gaurav:
Solved by PR #333 (I hope), so closing this PR.

PR #315, opened 2024-01-16
Display multiple phylogenies

gaurav:
This PR is a place for working on displaying multiple phyloreferences on a single phylogeny. It also implements some changes to how we display phyloreferences that resolve to a tip.

Initially, we don't display any phyloreferences in the summary.

Screenshot 2025-01-29 at 6 58 04 AM

After resolution, we can display phyloreferences in a table under the phylogeny next to the phylogeny node labels that it corresponds to. This also works for node labels like "Crocodylidae", which have two phyloreferences that resolve to it. Note that this in this display, we don't change any labels on the phylogeny at all.

Screenshot 2025-01-29 at 6 58 38 AM

We can export this phylogeny as an [annotated Nexus file](https://github.com/user-attachments/files/18587290/Alligatoridae_Alligatorinae_and_4_others.nex.txt), which can be opened in FigTree.

Screenshot 2025-01-29 at 7 00 29 AM

When displaying a single phyloreference, however, we do change phylogeny labels to indicate where the phyloreference has resolved, whether at an internal node (even with multiple phyloreferences):

Screenshot 2025-01-29 at 6 59 00 AM

Screenshot 2025-01-29 at 6 59 06 AM

Or at a tip:

Screenshot 2025-01-29 at 6 58 53 AM


gaurav:
- [ ] Turn phylorefs into links -- when you click on them, open the phyloref from the sidebar.

gaurav:
- Nothing wrong or misrepresented, but it can be misconstrued that this clade is only one node -- when really there is a whole clade back there.
- [ ] For branch-based, we do want to highlight the entire branch up to the previous node.
- [ ] For node-based, we don't want to do this.
- [ ] Reserve italics for either phyloreferences or internal labels without clade names.
- [ ] Change the internal node labels so they're in the same font as the terminal labels -- this will indicate that they come from the same place.
- [ ] For the individual phyloreference, don't display the internal label, but rather the phyloreference. Displaying the internal node labels for the summary makes more sense, but don't use font/weight/italics to distinguish them from the others, because otherwise people will wonder why they're different and whether they're coming from the same place.
- [ ] Use hover-over to show the phylorefs resolving to that node.

gaurav:
- [ ] Visualizing multiple phylorefs on a single node or edge in Nexus goes beyond phyloreferencing.

gaurav:
- [ ] For branch-based phyloreference, we could add a symbol to the clade, and then add a note underneath the phylogeny to say that this is clade extends up the parent branch.
  - Note that qualifiers can render a node-based clade into a branch-based clade, so we might want to look for "maximum" or other text in the definition.

gaurav:
Questions:
- [ ] Add table to display multiple resolution -- YES
- [ ] Support both FigTree and TreeViewer? -- NO, only FigTree. We do want multiple phyloreferences on a single node, which would require A||B concatenation.
- [ ] Do we still want to include expected resolution? -- NO, at least for now, that seems like enough.

src/components/phylogeny/Phylotree.vue
```diff
@@ -324,169 +359,213 @@ export default {
...
 
           // Wrap the phyloref is there is one.
-          const wrappedPhyloref = new PhylorefWrapper(this.phyloref || {});
+          this.phylorefs.forEach((phyloref) => {
+            const wrappedPhyloref = new PhylorefWrapper(phyloref || {});
 
-          // Make sure we don't already have an internal label node on this SVG node!
-          let textLabel = element.selectAll("text");
+            // Make sure we don't already have an internal label node on this SVG node!
+            let textLabel = element.selectAll("text");
```
hlapp, line 366: Are you sure the indentation in this and the preceding line is correct? It looks like there's extraneous space, but perhaps that's an artifact from tabs being used?

src/components/phylogeny/Phylotree.vue
```diff
@@ -324,169 +359,213 @@ export default {
...
+            // label it now!
+            if (
+              phyloref !== undefined &&
+              has(data, "@id") &&
+              this.$store.getters
+                .getResolvedNodesForPhylogeny(this.phylogeny, phyloref)
+                .includes(data["@id"])
+            ) {
+              // We found another pinning node!
+              if(!(data["@id"] in phylorefsByNodeId)) phylorefsByNodeId[data["@id"]] = new Set();
```
hlapp, line 453: Indent the statement under the conditional? Perhaps having it on the same line has been part of our coding style all along but it did surprise me here and took looking thrice to understand while scrolling through.

src/components/phylogeny/Phylotree.vue
```diff
@@ -324,169 +359,213 @@ export default {
...
+              element.classed("pinning-node", true);
+
+              // If there is no circle, add one.
+              if (element.select("circle").empty()) {
+                element.append("circle").attr("cx", -3).attr("r", 4);
               } else {
-                // Set the new label.
-                node.name = newName;
+                // Make the pinning node circle larger (slightly larger than its usual size of 3).
+                element.select("circle").attr("r", 4);
```
hlapp, line 467: Perhaps the usual size should be a constant, and then we're just incrementing that constant here by an absolute (+1) or relative (+33%) value?

src/components/phylogeny/Phylotree.vue
gaurav: That is odd, but yes, it looks like the alignment is correct:

https://github.com/phyloref/klados/blob/cbcf4ea8880ebc3cfb67f53fb82db14bd73db74d/src/components/phylogeny/Phylotree.vue#L363-L368

src/components/phylogeny/Phylotree.vue
gaurav: Done in d476dbd525cfbf3b6b397e252540afeed33d2dfd.

I haven't used our code linter in a while >.< Once we're done with the Klados 1.0 work, I should run it to see what can be cleaned up automatically/look into integrating that into PRs automatically.

gaurav:
> I think I've seen the result during our last meeting (or have I?) but regardless it would be useful to have updated screenshots of the visual results.

These screenshots are up to date! I'll go ahead and merge this PR and then make a Klados release so you can play with this at https://www.phyloref.org/klados/; please let me know if anything doesn't work as expected or if you'd like me to go through this at our next meeting.

Issue #257, opened 2022-08-22
Document CSV export format


gaurav:
The CSV export format is designed to mimic the summary table we show the user, but it would be good for downstream applications to have this format documented on the website or docs somewhere.

gaurav:
This should go into the Klados wiki, with a link to it from the application.

PR #321, opened 2024-02-28
Documentation and Summary page improvements

### LINKED ISSUE #168: Add support for documenting qualifiers in clade definitions
Currently, this can only be specified in the verbatim clade definition or curation notes. We should implement this as a type of specifier so that they can be specified using the same mechanism that allows you to choose a taxon/taxon concept, specimen, apomorphy or external reference.

### LINKED ISSUE #300: Improve the Phyx summary page
Two minor fixes may be needed to the Phyx summary page:

1. The phyloref table adds a column for each phylogeny, but currently only uses the label of the phylogeny (e.g. "Fig 1 from Brochu 2003"). Is this clear enough, or should we add "Phylogeny: " to the front of the phylogeny label (e.g. "Phylogeny: Fig 1 from Brochu 2003")?

- [x] If we change this, remember that our default name for unlabeled phylogenies is `Phylogeny 1`, so we want to make sure we don't end up writing this as `Phylogeny: Phylogeny 1`!

https://github.com/phyloref/klados/blob/ce4e31b1de251ff10a22cbc5b2e17ceb41f73436/src/components/phyx/PhyxView.vue#L105-L107

3. The phylogeny table refers to the "Description" of each phylogeny, but this field no longer exists. Should we replace this field with the curator notes? Or some other field, such as the source, number of leaf nodes, list of leaf nodes, etc.?

https://github.com/phyloref/klados/blob/ce4e31b1de251ff10a22cbc5b2e17ceb41f73436/src/components/phyx/PhyxView.vue#L214

Screenshot 2023-06-21 at 3 27 21 AM

gaurav:
- Added note that definition should include qualifiers (closes #168)
- Documented the CSV export format (related to #257, but I won't close that until I like to it from somewhere in the interface) as [a Markdown file](https://github.com/phyloref/klados/blob/bfae34e42e954920191e3e473f31ee2ae49ffdde/docs/ExportFormats.md).
- Fixed a bug in which we were providing all the expected resolutions before all the actual resolution in the CSV file, despite the header saying that they should be interleaved (i.e. `expected 1, actual 1, expected 2, actual 2`).
- The Open Tree of Life lookup code was previously recording the provenance of the phylogeny in the phylogeny description field, but since this field is no longer used, I've moved it into the phylogeny curator notes field (see example below).
- I accidentally turned on "reformat file on save", and so my IDE changed a bunch of file formatting without me noticing. That shouldn't change any functionality and should be easier to read/maintain in the future, but it's still annoying. 

This PR also closes #300 by modifying the summary table to make it clear that the final columns refer to phylogenies and replacing phylogeny descriptions (which are no longer displayed in Klados) with phylogeny curator notes.

Screenshot 2024-02-28 at 12 07 30 AM

Screenshot 2024-02-28 at 12 07 42 AM

### DIFF
--- a/docs/ExportFormats.md
+++ b/docs/ExportFormats.md
@@ -20,7 +20,7 @@ The summary table CSV export includes the following columns:
   * Example: `Caiman crocodilus`
 * `External specifier N`: A description of each external specifier (_N_ is a number starting from 1).
   * Example: `Caiman crocodilus`
-* `Phylogeny N: [Phylogeny label] expected`: The node resolved for this phylogeny (_N_ is a number starting from 1, and _Phylogeny label_ is the label of this phylogeny).
+* `Phylogeny: [Phylogeny label] expected`: The node resolved for this phylogeny (_N_ is a number starting from 1, and _Phylogeny label_ is the label of this phylogeny).
   * Example: `Alligatoridae`
-* `Phylogeny N: [Phylogeny label] actual`: The node resolved for this phylogeny (_N_ is a number starting from 1, and _Phylogeny label_ is the label of this phylogeny).
+* `Phylogeny: [Phylogeny label] actual`: The node resolved for this phylogeny (_N_ is a number starting from 1, and _Phylogeny label_ is the label of this phylogeny).
   * Example: `Alligatoridae_ott195670`, or `(unlabelled)` if the node has no label
--- a/docs/ExportFormats.md
+++ b/docs/ExportFormats.md
@@ -1,0 +1,26 @@
+# Export Formats
+
+## Summary table CSV export
+
+The summary table CSV export includes the following columns:
+
+* `Phyloreference ID`: A URL or URI (absolute, relative or local) that identifies this phyloreference.
+  * Example `#Alligatoridae`
+* `Label`: A label used to identify this phyloreference.
+  * Example: `Alligatoridae`
+* `Type`: The type of this phyloreference.
+  * May be one of the following values:
+    * `Apomorphy-based clade definition`
+    * `Maximum clade definition`
+    * `Minimum clade definition`
+    * `Invalid definition (...)` (with the reason for the invalid definition in brackets)
+* `Definition`: The clade definition as entered in the `Definition in free-form text`.
+  * Example: `Last common ancestor of Crocodylus niloticus, Osteolaemus tetraspis, and Tomistoma schlegelii and all of its descendents.`
+* `Internal specifier N`: A description of each internal specifier (_N_ is a number starting from 1).
+  * Example: `Caiman crocodilus`
+* `External specifier N`: A description of each external specifier (_N_ is a number starting from 1).
+  * Example: `Caiman crocodilus`
+* `Phylogeny N expected: [Phylogeny label]`: The node resolved for this phylogeny (_N_ is a number starting from 1, and _Phylogeny label_ is the label of this phylogeny).
+  * Example: `Alligatoridae`
+* `Phylogeny N actual: [Phylogeny label]`: The node resolved for this phylogeny (_N_ is a number starting from 1, and _Phylogeny label_ is the label of this phylogeny).
+  * Example: `(unlabelled)` if the node has no label
--- a/docs/ExportFormats.md
+++ b/docs/ExportFormats.md
@@ -20,7 +20,7 @@ The summary table CSV export includes the following columns:
   * Example: `Caiman crocodilus`
 * `External specifier N`: A description of each external specifier (_N_ is a number starting from 1).
   * Example: `Caiman crocodilus`
-* `Phylogeny: [Phylogeny label] expected`: The node in this phylogeny to which the phyloreference was expected to resolve (by the author) (_Phylogeny label_ is the label of this phylogeny).
+* `Phylogeny: [Phylogeny label] expected`: The label of the node in the specified phylogeny to which the phyloreference was expected to resolve by the author.
   * Example: `Alligatoridae`
-* `Phylogeny: [Phylogeny label] actual`: The node resolved for this phylogeny (_N_ is a number starting from 1, and _Phylogeny label_ is the label of this phylogeny).
+* `Phylogeny: [Phylogeny label] actual`: The label of the node actually resolved in the specified phylogeny to which the phyloreference was expected to resolve by the author.
   * Example: `Alligatoridae_ott195670`, or `(unlabelled)` if the node has no label
--- a/docs/ExportFormats.md
+++ b/docs/ExportFormats.md
@@ -20,7 +20,7 @@ The summary table CSV export includes the following columns:
   * Example: `Caiman crocodilus`
 * `External specifier N`: A description of each external specifier (_N_ is a number starting from 1).
   * Example: `Caiman crocodilus`
-* `Phylogeny: [Phylogeny label] expected`: The node resolved for this phylogeny (_N_ is a number starting from 1, and _Phylogeny label_ is the label of this phylogeny).
+* `Phylogeny: [Phylogeny label] expected`: The node in this phylogeny to which the phyloreference was expected to resolve (by the author) (_Phylogeny label_ is the label of this phylogeny).
   * Example: `Alligatoridae`
 * `Phylogeny: [Phylogeny label] actual`: The node resolved for this phylogeny (_N_ is a number starting from 1, and _Phylogeny label_ is the label of this phylogeny).
   * Example: `Alligatoridae_ott195670`, or `(unlabelled)` if the node has no label
--- a/docs/ExportFormats.md
+++ b/docs/ExportFormats.md
@@ -13,7 +13,7 @@ The summary table CSV export includes the following columns:
     * `Apomorphy-based clade definition`
     * `Maximum clade definition`
     * `Minimum clade definition`
-    * `Invalid definition (...)` (with the reason for the invalid definition in brackets)
+    * `Invalid definition (...)` (with the reason for the invalid definition in parentheses)
 * `Definition`: The (human readable) clade definition as entered in the `Definition in free-form text`.
   * Example: `Last common ancestor of Crocodylus niloticus, Osteolaemus tetraspis, and Tomistoma schlegelii and all of its descendents.`
 * `Internal specifier N`: A description of each internal specifier (_N_ is a number starting from 1).
--- a/docs/ExportFormats.md
+++ b/docs/ExportFormats.md
@@ -20,7 +20,7 @@ The summary table CSV export includes the following columns:
   * Example: `Caiman crocodilus`
 * `External specifier N`: A description of each external specifier (_N_ is a number starting from 1).
   * Example: `Caiman crocodilus`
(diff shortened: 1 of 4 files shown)


gaurav:
- [ ] No need to add "(including qualifiers)" -- free-form text already says I can write whatever I want.

docs/ExportFormats.md
```diff
@@ -0,0 +1,26 @@
+# Export Formats
+
+## Summary table CSV export
+
+The summary table CSV export includes the following columns:
+
+* `Phyloreference ID`: A URL or URI (absolute, relative or local) that identifies this phyloreference.
```
hlapp, line 7:
```suggestion
* `Phyloreference ID`: A URI (absolute, relative or local; may or may not resolve) that identifies this phyloreference. Normally intended for machine consumption.
```

hlapp: A URL is a URI, so it sounds weird to say _URL or URI_.

docs/ExportFormats.md
```diff
@@ -0,0 +1,26 @@
...
+  * Example `#Alligatoridae`
+* `Label`: A label used to identify this phyloreference.
```
hlapp, line 9:
```suggestion
* `Label`: A label for this phyloreference, normally intended for human consumption.
```

docs/ExportFormats.md
```diff
@@ -0,0 +1,26 @@
...
+  * Example: `Alligatoridae`
+* `Type`: The type of this phyloreference.
+  * May be one of the following values:
+    * `Apomorphy-based clade definition`
+    * `Maximum clade definition`
+    * `Minimum clade definition`
+    * `Invalid definition (...)` (with the reason for the invalid definition in brackets)
```
hlapp, line 16:
```suggestion
    * `Invalid definition (...)` (with the reason for the invalid definition in parentheses)
```

docs/ExportFormats.md
```diff
@@ -0,0 +1,26 @@
...
+* `Definition`: The clade definition as entered in the `Definition in free-form text`.
```
hlapp, line 17:
```suggestion
* `Definition`: The (human readable) clade definition as entered in the `Definition in free-form text`.
```

src/components/phyloref/PhylorefView.vue
```diff
@@ -51,7 +51,7 @@
               for="definition"
               class="col-form-label col-md-2"
             >
-              Definition in free-form text
+              Definition in free-form text (including qualifiers)
```
hlapp, line 54: I thought we were going to leave out the parenthetical text?

docs/ExportFormats.md
```diff
@@ -0,0 +1,26 @@
...
+  * Example: `Last common ancestor of Crocodylus niloticus, Osteolaemus tetraspis, and Tomistoma schlegelii and all of its descendents.`
+* `Internal specifier N`: A description of each internal specifier (_N_ is a number starting from 1).
+  * Example: `Caiman crocodilus`
+* `External specifier N`: A description of each external specifier (_N_ is a number starting from 1).
+  * Example: `Caiman crocodilus`
+* `Phylogeny: [Phylogeny label] expected`: The node resolved for this phylogeny (_N_ is a number starting from 1, and _Phylogeny label_ is the label of this phylogeny).
```
hlapp, line 23:
```suggestion
* `Phylogeny: [Phylogeny label] expected`: The node in this phylogeny to which the phyloreference was expected to resolve (by the author) (_Phylogeny label_ is the label of this phylogeny).
```

hlapp: There is no _N_ (nor should there be, right?), so I deleted the part about explaining _N_ (potentially a leftover from overzealous copy&paste?)

docs/ExportFormats.md
```diff
@@ -0,0 +1,26 @@
...
+  * Example: `Alligatoridae`
```
hlapp, line 24: This isn't compliant with the current item format description?

docs/ExportFormats.md
hlapp: And are you sure this should be the label of the phylogeny and not the node in the phylogeny?

docs/ExportFormats.md
```diff
@@ -0,0 +1,26 @@
...
+* `Phylogeny: [Phylogeny label] actual`: The node resolved for this phylogeny (_N_ is a number starting from 1, and _Phylogeny label_ is the label of this phylogeny).
```
hlapp, line 25: See above comments for _expected_, they apply here too.

hlapp (changes requested):
See inline notes. Perhaps a bit of a mess left from copy&paste?

src/components/phyloref/PhylorefView.vue
gaurav: Yes, I just hadn't gotten around to doing it yet! Fixed in 687dc04.

docs/ExportFormats.md
gaurav: Fixed in f9d2be48fed98f76e2a7e1491c0e6fcf9d348ffd.

docs/ExportFormats.md
gaurav: It's supposed to be the label of the node on the specified phylogeny where this phyloreference is expected to resolve -- so if phyloref #Alligatorinae is expected to resolve at the node "Alligatoridae" on the phylogeny "Fig 1 from Brochu 2003", the column will be named "Phylogeny: Fig 1 from Brochu 2003 expected" and will have the value "Alligatoridae". Does this make sense, or do I need to clarify something?

gaurav:
Thanks for the feedback, @hlapp! I've rewritten the "expected" and "actual" column descriptions as following:

* `Phylogeny: [Phylogeny label] expected`: The label of the node in the specified phylogeny to which the phyloreference was expected to resolve by the author.
  * Example: `Alligatoridae`
* `Phylogeny: [Phylogeny label] actual`: The label of the node actually resolved in the specified phylogeny to which the phyloreference was expected to resolve by the author.
  * Example: `Alligatoridae_ott195670`, or `(unlabelled)` if the node has no label

Is this clear?

The _N_ in these descriptions was a defunct description of what we do if a phylogeny is unlabeled; in this case, instead of labeling this column `Phylogeny: [Phylogeny label] expected`, we use `Phylogeny [Phylogeny index] expected`, e.g. `Phylogeny 1 expected`. Should we include this information in this documentation as well? We could also use `Phylogeny: [Phylogeny index] expected` so they're a bit more similar to each other. What do you think?

docs/ExportFormats.md
```diff
@@ -0,0 +1,26 @@
...
+* `Definition`: The (human readable) clade definition as entered in the `Definition in free-form text`.
+  * Example: `Last common ancestor of Crocodylus niloticus, Osteolaemus tetraspis, and Tomistoma schlegelii and all of its descendents.`
+* `Internal specifier N`: A description of each internal specifier (_N_ is a number starting from 1).
+  * Example: `Caiman crocodilus`
+* `External specifier N`: A description of each external specifier (_N_ is a number starting from 1).
+  * Example: `Caiman crocodilus`
+* `Phylogeny: [Phylogeny label] expected`: The label of the node in the specified phylogeny to which the phyloreference was expected to resolve by the author.
+  * Example: `Alligatoridae`
+* `Phylogeny: [Phylogeny label] actual`: The label of the node actually resolved in the specified phylogeny to which the phyloreference was expected to resolve by the author.
+  * Example: `Alligatoridae_ott195670`, or `(unlabelled)` if the node has no label
```
hlapp, line 26: Isn't it Tidy Data practice to have NA or an equivalent value representation to signify that there is no value? Why should we make up a special string to signify the absence of a value? Or is this the special value that's already present on the phylogeny?

hlapp:
See inline note. Can you comment, fix, and/or clarify?

docs/ExportFormats.md
gaurav: The idea behind `(unlabelled)` was to distinguish between the case where a phyloref doesn't resolve at all (I think we currently represent this as blank in the `Phylogeny: [Phylogeny label] actual` field) and the case where it does resolve, but it does so on a node that doesn't have a label. Some options:
- `NA` for unlabelled node and blank for no resolution
- `(unlabelled)` (or some other standard string) for unlabelled node and `NA` (or some other string, e.g. `(could not resolve)`, `None`) for no resolution
- `NA: unlabelled node` vs `NA: could not resolve`
- We could use the internal node identifier, which looks something like `#phylogeny0_node0`, if the resolved node is unlabelled and blank if the phyloref could not be resolved
- Add additional `Phylogeny: [Phylogeny label] error` column, which is blank unless the `... actual` column is `NA`.

What do you think?

hlapp: Perhaps this: _We could use the internal node identifier, which looks something like #phylogeny0_node0, if the resolved node is unlabelled and blank if the phyloref could not be resolved_?

But on reflection, using `(unlabelled)` and `(could not resolve)` seem at least as clear if not clearer.

hlapp:
I'm having trouble parsing _The label of the node actually resolved in the specified phylogeny to which the phyloreference was expected to resolve by the author_ in contrast to _The label of the node in the specified phylogeny to which the phyloreference was expected to resolve by the author_. In the first case, is the node to which the phyloreference is found to resolve different from what the author seems to have expected, or is it the node's label that different from what the author seemed to have expected?

docs/ExportFormats.md
gaurav: > But on reflection, using `(unlabelled)` and `(could not resolve)` seem at least as clear if not clearer.

Agreed! I've implemented this in 5464ea6b6012d92c90cfda15d328029feb8a6cef.

gaurav:
> I'm having trouble parsing _The label of the node actually resolved in the specified phylogeny to which the phyloreference was expected to resolve by the author_ in contrast to _The label of the node in the specified phylogeny to which the phyloreference was expected to resolve by the author_. In the first case, is the node to which the phyloreference is found to resolve different from what the author seems to have expected, or is it the node's label that different from what the author seemed to have expected?

Ugh, yes, that's a bit of a mess. How about:
* `Phylogeny: [Phylogeny label] expected`: The label of the node in the specified phylogeny to which the phyloreference was expected to resolve.
* `Phylogeny: [Phylogeny label] actual`: The label of the node actually resolved for this phyloreference in the specified phylogeny.

I've implemented this change in 2003205.

docs/ExportFormats.md
gaurav: I've made this a bit clearer in d9f349f:

* `Phylogeny: [Phylogeny label] expected`: The label of the node in the specified phylogeny to which the phyloreference was expected to resolve.
  * Example: `Alligatoridae`
  * Will be set to `(none)` if no expected node was found.
* `Phylogeny: [Phylogeny label] actual`: The label of the node actually resolved for this phyloreference in the specified phylogeny.
  * Example: `Alligatoridae_ott195670`
  * Will be set to `(resolution not yet run)` if resolution has not yet been run.
  * Will be set to `(could not resolve)` if this phyloreference could not be resolved.
  * Will be set to `(unlabelled)` if a node was resolved, but it had no label.

hlapp (approved):
Looks all good, finally 👍

Issue #12, opened 2018-03-27
Make selected UI elements more visible


gaurav:
Selected phyloreferences are not highlighted brightly enough -- make them really intensely coloured, and consider scrolling to them when they are selected. Remember that changing from black to any other colour reduces the constrast between them, so think about highlighting their background as well.

The idea here seems to be that my eye/screen setup is better than most, so I should make it really obvious and garish on my screen, and then it'll be visible for them.

gaurav:
In 33dbd8f, I've moved the node descriptions into CSS so it's easier to edit them. Currently, I'm trying:
- Nodes: 10pt
- Internal specifier node: 12pt green bolder
- External specifier node: 12pt magenta bolder
- Terminal nodes without taxonomic units: turkey red italic bolder
- Internal label: 12pt serif italic
- Selected internal label: 16pt blue

gaurav:
Increasing the font size and changing the font colour don't seem to have helped very much. I'll try adding a coloured (light-grey) rectangle behind the text as a highlight, as in https://stackoverflow.com/a/16873128/27310

gaurav:
- [ ] When a clade resolves to a terminal node, we should make the font size larger (the same size as the clade label in the tree) so as to make it clear that this node was resolved as the entire clade.

gaurav:
- [ ] If it's a terminal node, we should use the phyloreference name rather than the terminal name (and them maybe add the terminal name afterwards). e.g. `Caimaninae (Caiman crocodylus)`

gaurav:
- [ ] Don't need a legend: it is somewhat clear from the context why certain things are highlighted.

gaurav:
- [x] Keep the new labeled node display.
- [ ] If a phyloref expects to resolve to a node label that doesn't exist, we should highlight that (we do that for non-existent nodes, but this doesn't work for phyloref nodes that expect to resolve to an alternate label)
- [ ] Optionally display all node identifiers? Maybe on hover? Worst case, we can put it in the popup menu.
- [ ] Change "Go to node" to "Scroll to node".

gaurav:
- [x] Stick with the new colours.

PR #322, opened 2024-02-28
Improve display of phyloref resolution on phylogeny

gaurav:
WIP

Relates to #12.


gaurav:
It would be better to tackle this as part of PR #315, since displaying it all together would be a good way to clarify what we need. Closing for now.

PR #323, opened 2024-02-28
Replaced Vue variables with Vite variables, including APP_VERSION

### LINKED ISSUE #316: Code for importing Git VERSION into Klados needs to be updated
We used to have a system for setting the `VUE_APP_VERSION` environmental variable from the GitHub Action that builds the website, so that the website version would be included in the header of the app.

https://github.com/phyloref/klados/blob/1e576d21b80b24dbdf722d524cc1c15a8928b2e6/src/App.vue#L63-L65

However, Vite uses a different system for storing and exposing environmental variables (https://vitejs.dev/guide/env-and-mode.html) and so this system will need to be changed.

gaurav:
Vue variables have not been working since we switched over to use Vite as our Vue runner, which is why https://www.phyloref.org/klados/ says "(unversioned)" as its version number. This PR should fix that.

Closes #316.

### DIFF
--- a/.github/workflows/deploy-to-github-pages.yml
+++ b/.github/workflows/deploy-to-github-pages.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Build Vue page in /docs
         run: |
           npm install
-          VUE_APP_VERSION=$GITHUB_REF_NAME npm run build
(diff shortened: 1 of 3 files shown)


hlapp (approved):
👍

PR #329, opened 2024-04-24
Improve double quote warning for Nexus export

### LINKED ISSUE #328: Warning text upon export to NEXUS when an annotation already contains single quotes
@gaurav has implemented the suggested improvements in 93b09e3 -- we now convert double quotes to single quotes, but if an annotation already contains single quotes, we produce a warning in a single confirmation dialog that looks like this:

Screenshot 2024-04-10 at 12 09 13 AM

Clicking on "Cancel" cancels the export.

The dialog box says:

> Some annotations (such as ${annotations_containing_single_and_double_quotes.slice(0, 3).join("; ")}) contain both single quotes and double quotes. Double quotes need to be converted to single quotes so that the resulting NEXUS file can be read, but this will be non-reversible. Do you want to export this NEXUS file?

I'll go ahead and merge this PR since it's been approved, but please open a ticket if we need to modify this text.

_Originally posted by @gaurav in https://github.com/phyloref/klados/issues/313#issuecomment-2046509469_

gaurav:
Closes #328.

Example dialog when trying to export a phylogeny annotated with a phyloreference labeled `Alligatoridae's "real" way`:
Screenshot 2024-04-24 at 3 03 30 AM

### DIFF
--- a/src/components/phylogeny/Phylotree.vue
+++ b/src/components/phylogeny/Phylotree.vue
@@ -240,9 +240,9 @@ export default {
           !window.confirm(
             `Some annotations (such as ${example_annotations.join(
               "; "
-            )}) contain both single quotes and double quotes. Double quotes need to be converted to single quotes ` +
-              "so that the resulting NEXUS file can be read, but this will be non-reversible. Do you want to export " +
-              "this NEXUS file?"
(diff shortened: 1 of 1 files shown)


hlapp (approved):
Sorry for dropping this from my radar. All good 👍

PR #332, opened 2024-07-27
Fix uptime test

gaurav:
The [backend test](https://github.com/phyloref/klados/actions/workflows/test-backend.yml) have been failing because the certificate provided by https://reasoner.phyloref.org/reason isn't recognized by our backend job. I've added that certificate chain to this repository as a secret, which this PR reads into the test-backend script to use as a certificate when making the HTTPS request to the Reasoner backend.


hlapp (approved):
👍

PR #342, opened 2024-12-23
Update to @phyloref/phyx 0.2.0-alpha3

gaurav:
This PR is to test [@phyloref/phyx.js v1.2.0-alpha3](https://github.com/phyloref/phyx.js/pull/142), which has some improvements we'd like to get into Klados. We should test it here before releasing it as v1.2.0 to make sure it works well with Klados.


gaurav:
We already updated the `@phyloref/phyx.js` package in another PR, so this is no longer needed. Closing.

PR #343, opened 2025-01-15
Add Selenium tests to Klados

gaurav:
This PR will start the process of adding Selenium tests to Klados, so we can make sure that the core functionality of Klados doesn't break in future changes.

WIP


hlapp:
@gaurav FYI, in other projects we have standardized on [AGENTS.md](https://agents.md) instead of using Copilot or Claude-specific instructions.

PR #349, opened 2025-02-26
Make sure that the incoming Phyx file has both `phylorefs` and `phylogenies` keys

### LINKED ISSUE #330: "Add OTEL" button silently fails if the Phyx file doesn't have a `phylogenies` key
e.g. try clicking it after choosing the "Testudinata" example.

gaurav:
Klados acts in odd ways if the input file doesn't have both `phylorefs` and `phylogenies` keys (e.g. #330). This PR makes sure that the incoming Phyx file has both `phylorefs` and `phylogenies` keys.

Close #330.

### DIFF
--- a/src/components/sidebar/Sidebar.vue
+++ b/src/components/sidebar/Sidebar.vue
@@ -328,6 +328,13 @@ export default {
 
       $.getJSON(url)
         .done((data) => {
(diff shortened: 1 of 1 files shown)


hlapp (approved):
👍

PR #351, opened 2025-03-11
Citation should be stored in `namePublishedIn`, not `dwc:namePublishedIn`

### LINKED ISSUE #339: Is this field really `dwc:namePublishedIn` or should it be `namePublishedIn`?
https://github.com/phyloref/klados/blob/f397d027b61e41538a9062dcff533cd61ffee527/src/components/phyloref/PhylorefView.vue#L45

gaurav:
It is `namePublishedIn` in both Phyx publication (https://peerj.com/articles/12618/#table-5) and in both v1.0.0 and v1.1.0 of the phyx.js context (https://github.com/search?q=repo%3Aphyloref%2Fphyx.js%20namePublishedIn&type=code), so we should use that instead. Closes #339.

### DIFF
--- a/src/components/phyloref/PhylorefView.vue
+++ b/src/components/phyloref/PhylorefView.vue
@@ -42,7 +42,7 @@
           <Citation
             label="Name published in"
             :object="selectedPhyloref"
-            citation-key="dwc:namePublishedIn"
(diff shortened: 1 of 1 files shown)


hlapp (approved):
👍

PR #353, opened 2025-03-11
Improved placeholder text for the phylogeny curator notes field

### LINKED ISSUE #70: Make it clear that Reference Phylogeny information can be added to the Curator Notes field
Phyloreference definitions often contain a lot of useful information in the "Reference Phylogeny" section: apart from a citation to a particular phylogeny, they can also refer to other phylogenies, how they interpret features of a particular phylogeny, or point out missing specifiers and where they believe them to be placed. This should be added as an additional field below the clade definition.

gaurav:
This PR improves the placeholder text for the phylogeny curator notes field. Closes #70.

This is what the placeholder looks like now:
Screenshot 2025-03-11 at 1 17 06 AM

Note that the other placeholders (e.g. for phylorefs) aren't changed, but this would be a good PR to change them if needed:

Screenshot 2025-03-11 at 1 16 58 AM

### DIFF
--- a/src/components/phylogeny/PhylogenyView.vue
+++ b/src/components/phylogeny/PhylogenyView.vue
@@ -68,7 +68,7 @@
                 v-model.trim="phylogenyCuratorNotes"
                 type="text"
                 class="form-control"
-                placeholder="Enter curator notes for this phylogeny, such as any differences from its published form."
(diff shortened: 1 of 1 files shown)


hlapp (approved):
👍

PR #361, opened 2025-04-23
Make selected UI more visible

### LINKED ISSUE #12: Make selected UI elements more visible

gaurav:
- Changed the color for internal specifier text, external specifier text and selected internal node. 
- Changed button text on phylogeny expected/actual resolution labels from "Go to node" to "Scroll to node".
- Replaced `getResolvedNodes()` with a `getResolvedNodeLabels()` method duplicated from PhyxView (where it is called `getNodeLabelsResolvedByPhyloref()`). I don't like the duplication, but the code is complicated enough that I don't want to try consolidating it yet -- I've left a TODO for us to fix this later.
- Added both the node ID and the node name to the popup menu when left-clicking on a node. These menu items can't be clicked (doing so just closes the popup menu), and unfortunately custom menu items are always placed at the bottom. But I think it could be useful.
- Added an option for displaying every node ID as hover text over the node. For now, I've turned this on, but we can turn it off if we need to.
- Updated the example Phyx files (Brochu 2003 with an absolute URL for the phylogeny, Fisher et al 2007 with the updated full representation of each taxon name).
- Confirmed that an error message is displayed if the expected node is not present in the phylogeny.
- Turn the button below Phylotree into a Button Group -- this will make it easier to add more buttons in the future if necessary.

Closes #12.

Display if the expected node is not found in a phylogeny:
Screenshot 2025-04-22 at 9 11 57 PM

## If phylogeny URL is a relative URL:
Popup menu displaying the node ID and node name:
Screenshot 2025-04-22 at 9 09 42 PM

Node ID being displayed when hovering over it:
Screenshot 2025-04-22 at 9 09 32 PM

## If phylogeny URL is an absolute URL:
Popup menu displaying the node ID and node name:
Screenshot 2025-04-22 at 9 28 04 PM

Node ID being displayed when hovering over it:
Screenshot 2025-04-22 at 9 28 23 PM

### DIFF
--- a/src/components/phylogeny/Phylotree.vue
+++ b/src/components/phylogeny/Phylotree.vue
@@ -430,6 +430,20 @@ export default {
             // Clear any existing menu items.
             node.menu_items = [];
 
+            // Add a custom menu items to display the node label and ID.
+            addCustomMenu(
+              node,
+              (node) => "Node ID: " + (node.data['@id'] || "(none)"),
+              () => false,
+              (node) => true,
+            );
+            addCustomMenu(
+              node,
+              (node) => "Node label: " + (node.data.name || "(none)"),
+              () => false,
+              (node) => true,
+            );
+
             // Add a custom menu item to allow us to rename this node.
             // console.log("node", node);
             addCustomMenu(
--- a/src/components/phylogeny/Phylotree.vue
+++ b/src/components/phylogeny/Phylotree.vue
@@ -16,16 +16,18 @@
     <div v-else class="phylotreeContainer">
       <div :id="'phylogeny' + phylogenyIndex" class="col-md-12 phylogeny" />
       <ResizeObserver @notify="redrawTree" />
-      <button
-        type="button"
-        class="btn btn-primary"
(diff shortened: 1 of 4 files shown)


src/components/phylogeny/Phylotree.vue
```diff
@@ -416,6 +430,20 @@ export default {
             // Clear any existing menu items.
             node.menu_items = [];
 
+            // Add a custom menu items to display the node label and ID.
```
hlapp, line 433:
```suggestion
            // Add custom menu items to display the node label and ID.
```

hlapp (approved):
See inline suggestion (prompted by Copilot's comment).

PR #362, opened 2025-04-23
Make phyloref identifier editable

### LINKED ISSUE #357: Do we need to make the phyloref identifier editable?
Seems like that would be useful to have (phylogeny identifiers are already editable).

gaurav:
This PR makes the Phyloref identifier editable. Since this is not referenced elsewhere in the file, this should be safe to do without modifying anything else in the file. I've also added links to the ["URI" wikipedia article](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier). Unlike the phylogeny ID, this is not necessary (we make one up when we export to OWL) and should not be referenced from elsewhere in the Phyx file, so all we need to do is edit the field. 

Also, since a phylogeny ID is needed in some cases, I've added a check to make sure that it can't be set to blank.

Closes #357.

Phylogeny view:
Screenshot 2025-04-22 at 10 04 09 PM

Phylogeny view when empty:
Screenshot 2025-04-22 at 10 11 03 PM

Phyloref view:
Screenshot 2025-04-22 at 10 04 46 PM

Phyloref view when empty:
Screenshot 2025-04-22 at 10 09 10 PM

Error if a duplicate ID is provided:
Screenshot 2025-04-22 at 10 04 38 PM
Screenshot 2025-04-22 at 10 04 24 PM

### DIFF
--- a/src/components/phylogeny/PhylogenyView.vue
+++ b/src/components/phylogeny/PhylogenyView.vue
@@ -18,12 +18,12 @@
               for="phylogenyId"
               class="col-md-2 col-form-label"
             >
-              Identifier
+              Phylogeny ID (should be a <a target="_blank" href="https://en.wikipedia.org/wiki/Uniform_Resource_Identifier">URI</a>)
             </label>
             <div class="col-md-10">
               <input
                 id="phylogenyId"
-                v-model="phylogenyId"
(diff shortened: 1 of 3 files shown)


hlapp (approved):
👍

PR #363, opened 2025-04-23
Add support for concatenating Phyx files

### LINKED ISSUE #319: Add support for adding a Phyx file to an existing Phyx file
Problem statement: A user has two Phyx files that they want to simultaneously visualize: either they wish to build a single ontology containing phyloreferences from both files, or they wish to resolve phyloreferences from one Phyx file against phylogenies from another file.

User interface implementation:
1. Add a "Add Phyx file to current file" button.
2. On clicking this button, we ask the user to choose a Phyx file to add.
3. All phyloreferences and phylogenies from the second Phyx file are added to the in-memory Phyx file.

Merging two Phyx files in this way should be straightforward but should be implemented in phyx.js.

gaurav:
This PR will add a menu item called "Concatenate local JSON file" that will allow users to load a local Phyx file. As long as no phylogeny or phyloref ID in the second Phyx document overlaps with the current one, we should be able to just concatenate the phylorefs and phylogenies to the existing list. We won't modify file contents outside of those fields, e.g. the default nomenclatural code and other file-level properties won't be changed. Closes #319.

WIP. Testing this appears to have revealed a problem with the phylogeny ID changing code, so I'll test that a bit more thoroughly before wrapping this up.

### DIFF
--- a/src/components/sidebar/Sidebar.vue
+++ b/src/components/sidebar/Sidebar.vue
@@ -23,7 +23,21 @@
           type="file"
           class="d-none"
           @change="loadPhyxFromFileInputById('#file-input')"
+        />
+
+        <a
+          class="list-group-item list-group-item-action"
+          href="javascript: void(0)"
+          onclick="$('#file-input-concat').trigger('click')"
         >
+          Concatenate local JSON file
+        </a>
+        <input
+          id="file-input-concat"
+          type="file"
+          class="d-none"
+          @change="concatPhyxFromFileInputById('#file-input-concat')"
+        />
 
         <a
           class="list-group-item list-group-item-action"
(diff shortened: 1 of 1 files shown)


src/components/sidebar/Sidebar.vue
```diff
@@ -401,6 +415,107 @@ export default {
...
+        const phylorefsToAdd = newPhyx.phylorefs || [];
+        const phylogeniesToAdd = newPhyx.phylogenies || [];
+
+        // Step 1. Make sure we don't have any phylorefs with the same @id.
+        const currentPhylorefIds = (currentPhyx.phylorefs || [])
+          .map(phyloref => phyloref['@id'] || outerStore.getters.getPhylorefId(phyloref));
+        const phylorefIdsToAdd = phylorefsToAdd.map(phyloref => phyloref['@id'] || outerStore.getters.getPhylorefId(phyloref));
+        const phylorefIdsInCommon = currentPhylorefIds.filter(phylorefId => phylorefIdsToAdd.includes(phylorefId));
+        if (phylorefIdsInCommon.length > 0) {
+          alert('Cannot concatenate Phyx files -- the following phyloref IDs are present in the current file: ' + phylorefIdsInCommon.join(', '));
```
hlapp, line 488:
```suggestion
          alert('Cannot concatenate Phyx files -- the following phyloref IDs are already present in the current file: ' + phylorefIdsInCommon.join(', '));
```

src/components/sidebar/Sidebar.vue
```diff
@@ -401,6 +415,107 @@ export default {
...
+          return;
+        }
+
+        // Step 2. Make sure we don't have any phylogenies with the same @id.
+        const currentPhylogenyIds = (currentPhyx.phylogenies || [])
+          .map(phylogeny => phylogeny['@id'] || outerStore.getters.getPhylogenyId(phylogeny));
+        const phylogenyIdsToAdd = phylogeniesToAdd.map(phylogeny => phylogeny['@id'] || outerStore.getters.getPhylogenyId(phylogeny));
+        const phylogenyIdsInCommon = currentPhylogenyIds.filter(phylogenyId => phylogenyIdsToAdd.includes(phylogenyId));
+        if (phylogenyIdsInCommon.length > 0) {
+          alert('Cannot concatenate Phyx files -- the following phylogeny IDs are present in the current file: ' + phylogenyIdsInCommon.join(', '));
```
hlapp, line 498:
```suggestion
          alert('Cannot concatenate Phyx files -- the following phylogeny IDs are already present in the current file: ' + phylogenyIdsInCommon.join(', '));
```

hlapp (approved):
Looks good, except as perhaps a nitpick of sorts, the term "concatenate" causes a semantic confusion for me and I think "append" would be better and more accurately describe what's happening.

To me, _concatenate_ applies to a list of two or more things, whereas here we are only looking for a single Phyx file the contents of which will then be appended to the Phyx file we're already working on. (And in a comment you say _concatenate into_ which in my book really doesn't exist; instead we _insert into_ or _append to_.)

Having said that, it's relatively minor, appears in many places such as comments and function names, and as a change could well be postponed to later, so I'm approving this nonetheless. Feel free to decide whether to correct this or carve it out into its own issue to be addresses in the future.

